PR #1, opened 2017-11-14
Enable local docker image build

benjefferies:
The Dockerfile expected the jar location to be passed in as an extra
parameter making the builds not work as easily out of the box. Maven
will always build jars to the same locations so we can expect the jar to
be there.

Adding spotify docker plugin will allow us to hook into the maven build
to produce a docker image when running `mvn install`. The overhead of
creating the docker image when running the build is around 14s.

Implemented following the spring boot guide:
https://spring.io/guides/gs/spring-boot-docker/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,6 @@
-FROM openjdk:8u121-jre 
-MAINTAINER Kieran Wardle <kieran.wardle@ons.gov.uk>
-ARG jar
-VOLUME /tmp
-COPY $jar casesvc.jar
-RUN sh -c 'touch /casesvc.jar'
-ENV JAVA_OPTS=""
-ENTRYPOINT [ "sh", "-c", "java -jar /casesvc.jar" ]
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,9 +1,6 @@
...
-ARG jar
-VOLUME /tmp
-COPY $jar casesvc.jar
-RUN sh -c 'touch /casesvc.jar'
-ENV JAVA_OPTS=""
-ENTRYPOINT [ "sh", "-c", "java -jar /casesvc.jar" ]
+FROM openjdk:8-jre
+
+COPY target/casesvc*.jar /opt/casesvc.jar
+
```
madeye-matt: There is an edge case here - if you run 'mvn install' with no 'clean' after changing the pom version you'll end up with multiple files matching casesvc* - what's the docker behaviour in this case? Not sure there's a way around it though ...

PR #4, opened 2017-11-22
Docker java debug

madeye-matt:
Open port 5171 on docker container for debugging


jwgwalton:
Are we going to put something in the README about passing the debug environment variables, or would that all come from the compose?

benjefferies:
Might be worth adding a docker section to the readme? Maybe something that describes how to run the service with docker and the argument that can be added

PR #6, opened 2017-12-21
Changes for maven release plugin

benjefferies:
## What is it
These changes are driven to allow us to release artifacts in a much simpler way. Currently we have a bunch of shell scripts that we maintain to do releases when this is a core part of maven to be able to bump version, create tags and deploy artifacts.

* Bump versions and remove snapshot 
Release 13 was done on a branch where the version was incremented. The version change hasn't made it's way back into master yet so this change skips that version. We are also removing the snapshot versions.

* Using optional dockerfile buildArgs
We have been fighting with the `COPY target/casesvc*.jar` syntax when
there is  more that one casesvc*.jar file that  exists in the target
directory. Thought it was fixed by the docker info change but the same
issue occurs when doing a release because it builds the samplesvc-sources.jar.
This change makes the dockerfile JAR_FILE arg optional to allow the maven
build to provide the full name including version.

The output of the artifacts/tags are:
* A tag called `${project.artifactId}-${project.version}
* 3 artifacts called `${project.artifactId}-${project.version}-${classifier}.jar
* A version of the pom.xml in artifactory in the format `${project.artifactId}-${project.version}.pom`
## Testing
1. Checkout branch
2.  `export ARTIFACTORY_PASSWORD=$REAL_PASSWORD`
3. `mvn -B release:prepare release:perform -Darguments="-Dmaven.javadoc.skip=true"`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,6 @@
-ARG JAR_FILE=casesvc*.jar
 FROM openjdk:8-jre
 
-ARG JAR_FILE
+ARG JAR_FILE=casesvc*.jar
 COPY target/$JAR_FILE /opt/casesvc.jar
 
 ENTRYPOINT [ "sh", "-c", "java $JAVA_OPTS -jar /opt/casesvc.jar" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,8 @@
+ARG JAR_FILE=casesvc*.jar
 FROM openjdk:8-jre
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,6 +1,8 @@
+ARG JAR_FILE=casesvc*.jar
```
LJBabbage: Would be clearer if not before the FROM

PR #15, opened 2018-02-15
Release 15 hot fix

rjweeks:
Allow transition from NOT STARTED to COMPLETE.


LJBabbage:
Great quick fix, Thanks Richard

PR #19, opened 2018-03-06
Us051 generate new enrolment code

andrewmil:
Added new endpoint which creates a new case with a new enrolment code for a given collection_exercise_id and ru_ref

To test, run with other PR's
[response-operations-ui](https://github.com/ONSdigital/response-operations-ui/pull/85)
[ras-backstage-service](https://github.com/ONSdigital/ras-backstage/pull/71)
[rasrm-acceptance-tests](https://github.com/ONSdigital/rasrm-acceptance-tests/pull/68)

[Trello](https://trello.com/c/FIBrMsOb/72-us051-int-generate-a-new-additional-enrolment-code-l-25pts-5d)
[Confluence](https://digitaleq.atlassian.net/wiki/spaces/RASB/pages/227115065/US051+Provide+a+new+Enrollment+Code+to+an+RU+for+a+specific+survey+over+the+phone+Large)

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseEndpoint.java
+++ b/src/main/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseEndpoint.java
@@ -22,6 +22,7 @@ import uk.gov.ons.ctp.response.casesvc.domain.model.Case;
 import uk.gov.ons.ctp.response.casesvc.domain.model.CaseEvent;
 import uk.gov.ons.ctp.response.casesvc.domain.model.CaseGroup;
 import uk.gov.ons.ctp.response.casesvc.domain.model.Category;
+import uk.gov.ons.ctp.response.casesvc.message.sampleunitnotification.SampleUnitBase;
 import uk.gov.ons.ctp.response.casesvc.representation.CaseDTO;
 import uk.gov.ons.ctp.response.casesvc.representation.CaseDetailsDTO;
 import uk.gov.ons.ctp.response.casesvc.representation.CaseEventCreationRequestDTO;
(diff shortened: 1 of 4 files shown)


andrewmil:
[US051: INT Generate a new / additional enrolment code (L = 25pts = 5d)](https://trello.com/c/FIBrMsOb/72-us051-int-generate-a-new-additional-enrolment-code-l-25pts-5d)

src/main/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseEndpoint.java
```diff
@@ -151,6 +156,40 @@
...
+                      collectionexerciseid, ruref));
+    }
+    List<Case> casesList = caseService.findCasesByCaseGroupFK(caseGroup.getCaseGroupPK());
+    Case latestCase = casesList.get(casesList.size() - 1);
+    List<String> iacs = internetAccessCodeSvcClientService.generateIACs(1);
+
+    SampleUnitBase sampleUnitBase = new SampleUnitBase(caseGroup.getSampleUnitRef(), Character.toString('B'),
+            caseGroup.getPartyId().toString(), latestCase.getCollectionInstrumentId().toString());
+    Case newCase = caseService.generateNewCase(sampleUnitBase, caseGroup, iacs.get(0), latestCase.getActionPlanId());
+    if (newCase == null) {
```
jwgwalton: newCase will never be null. If there is an error when saving to the database it would come from a DataAccessException. https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/dao/DataAccessException.html

andrewmil:
@tej99 There may be cases where there are no ACTIONABLE B cases available but we will still want to generate an enrolment code. I think this happens any time there's no active enrolment codes but we want to generate one?

andrewmil:
@benjefferies I agree that we probably should be doing some kind of auditing, however right now we're only going to have one user in uaa so until that is not the case we can't really audit anything?

PR #24, opened 2018-03-22
Change eventtype for RESPONDENT_ENROLLED to disabled

jwgwalton:
So the disabling of the IAC is handled by the case event.


jwgwalton:
[TBL131: Add Survey deactivate IAC code should be done through a new case event (M)](https://trello.com/c/w2mZeXJv/124-tbl131-add-survey-deactivate-iac-code-should-be-done-through-a-new-case-event-m)

jwgwalton:
[TBL131: Add Survey deactivate IAC code should be done through a new case event (M)](https://trello.com/c/w2mZeXJv/124-tbl131-add-survey-deactivate-iac-code-should-be-done-through-a-new-case-event-m)

jwgwalton:
@benjefferies , there are a lot of tests based on disabled not deactivating the IAC code. That seems to be the only difference but a lot of them seem to be census specific.

jwgwalton:
@benjefferies for example
testCreatePaperResponseEventAgainstActionableCase specifically states that the iac is not disabled. 

    // IAC should not be disabled for paper responses
    verify(internetAccessCodeSvcClientService, times(0)).disableIAC(any(String.class));

I'd be more than happy to remove it as logically I don't see why you'd want an inactionable case without disabling the IAC. It does seem to have a specific use though.

PR #29, opened 2018-04-17
Performance Improvement

VibhorMehra:
- Changing `case-distribution_delay-milli-seconds` to 500. This is
required to speed up collectionExercise execution.

- Changing `createCaseEvent` logging to INFO. This is required to
investigate receipting issues without switching to DEBUG level.

- Changing `case-distribution_delay-milli-seconds` to 500 generates a
lot of noise in logs at DEBUG level. Hence, currently it is not recommended to leave
logging at DEBUG for a long time.


VibhorMehra:
[Performance: Performance test RAS/RM Release 16](https://trello.com/c/Rq4Dmaze/108-performance-performance-test-ras-rm-release-16)

PR #33, opened 2018-04-30
Fix link new b cases to business

gemmai95:
This is a fix for disable respondent enrolment when the respondent is the last actionable user for a case group a new B case is created so the business gets a letter for someone to enrol for a survey for that business. The current implementation links the new B case to a respondent rather than a business. Due to this the new enrolment codes do not work. This will fix this.

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/casesvc/service/impl/CaseServiceImpl.java
+++ b/src/main/java/uk/gov/ons/ctp/response/casesvc/service/impl/CaseServiceImpl.java
@@ -410,7 +410,6 @@ public class CaseServiceImpl implements CaseService {
         newCase.setCollectionInstrumentId(targetCase.getCollectionInstrumentId());
       }
 
-      log.info(newCase.getPartyId().toString(), '2');
       createNewCaseFromEvent(caseEvent, targetCase, newCase, category);
     }
   }
@@ -424,7 +423,6 @@ public class CaseServiceImpl implements CaseService {
    */
   private void buildNewCase(Category category, Case newCase, Case targetCase) {
 
-    log.info(newCase.getPartyId().toString(), '3');
     newCase.setSampleUnitType(SampleUnitType.valueOf(category.getNewCaseSampleUnitType()));
 
     // set case group id to the same as
@@ -441,7 +439,6 @@ public class CaseServiceImpl implements CaseService {
       }
     }
 
-    log.info(newCase.getPartyId().toString(), '4');
   }
 
   /**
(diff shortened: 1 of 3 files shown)


gemmai95:
[P2 Re-creating the B case when disabling the enrolment for the last enrolled BI case is assigning the respondent's partyid. It should be assigning the businesses partyid](https://trello.com/c/GjqggMzb/79-p2-re-creating-the-b-case-when-disabling-the-enrolment-for-the-last-enrolled-bi-case-is-assigning-the-respondents-partyid-it-sho)

src/main/java/uk/gov/ons/ctp/response/casesvc/service/impl/CaseServiceImpl.java
```diff
@@ -415,6 +422,7 @@ private void createNewCase(Category category, CaseEvent caseEvent, Case targetCa
    * @param newCase the new case to be created
    */
   private void buildNewCase(Category category, Case newCase, Case targetCase) {
+
```
mfcrocker: Would remove this newline

src/main/java/uk/gov/ons/ctp/response/casesvc/service/impl/CaseServiceImpl.java
```diff
@@ -430,6 +438,7 @@ private void buildNewCase(Category category, Case newCase, Case targetCase) {
         newCase.setActionPlanId(caseType.getActionPlanId());
       }
     }
+
```
mfcrocker: Would remove this newline

src/main/java/uk/gov/ons/ctp/response/casesvc/service/impl/CaseServiceImpl.java
```diff
@@ -219,6 +219,13 @@ else if (caseEvent.getCategory().equals(CategoryDTO.CategoryName.DISABLE_RESPOND
         }
       }
 
+      else if (caseEvent.getCategory().equals(CategoryDTO.CategoryName.REPLACED) && category.getNewCaseSampleUnitType().equals(SampleUnitType.B))  {
```
mfcrocker: A good refactoring option here would be changing this entire if/else if/else if block to a `switch(caseEvent.getCategory())`. This particular line is the only case that would cause a problem in doing this refactoring and can be simply changed as follows:

`case REPLACED:
if category.getNewCaseSampleUnitType().equals(SampleUnitType.B)) {
 // do stuff
}`

src/main/java/uk/gov/ons/ctp/response/casesvc/service/impl/CaseServiceImpl.java
```diff
@@ -219,6 +219,13 @@ else if (caseEvent.getCategory().equals(CategoryDTO.CategoryName.DISABLE_RESPOND
...
+          CaseGroup caseGroup = caseGroupRepo.findById(targetCase.getCaseGroupId());
```
mfcrocker: This variable is only used in the line below; I'd suggest getting rid of it and just going `newCase.setPartyId((caseGroupRepo.findById(targetCase.getCaseGroupId)).getPartyId());`

PR #35, opened 2018-05-08
Defect: No comms when respondent's enrolment is disabled after generating new enrolment code

andrewmil:
Previously, when a respondent enrolment was disabled, we were only creating a new B case when there were no ACTIONABLE cases for the given casegroups
We actually want to create this new case when there are no ACTIONABLE BI cases

[Trello](https://trello.com/c/qPFjHpuH/195-defect-ru-respondent-does-not-receive-any-comms-when-respondents-enrolment-is-disabled-after-generating-new-enrolment-code)


andrewmil:
[Defect: RU/Respondent does not receive any comms when respondent's enrolment is disabled after generating new enrolment code](https://trello.com/c/qPFjHpuH/195-defect-ru-respondent-does-not-receive-any-comms-when-respondents-enrolment-is-disabled-after-generating-new-enrolment-code)

PR #37, opened 2018-05-15
Remove source case Id for B cases being created

gemmai95:
In the action service if the source case for a B case is not null it is linked back to a BI case. This will fix this so all B cases should not have a source case. Tested with the disable enrolment story as this is where the defect was found.

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/casesvc/service/impl/CaseServiceImpl.java
+++ b/src/main/java/uk/gov/ons/ctp/response/casesvc/service/impl/CaseServiceImpl.java
@@ -2,7 +2,6 @@ package uk.gov.ons.ctp.response.casesvc.service.impl;
 
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.collections.CollectionUtils;
-import org.eclipse.jdt.core.dom.NullLiteral;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Propagation;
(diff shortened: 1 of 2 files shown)


gemmai95:
[When creating a B case for disable enrolment the new case is being linked to the BI case](https://trello.com/c/SqDKb9HG/91-when-creating-a-b-case-for-disable-enrolment-the-new-case-is-being-linked-to-the-bi-case)

src/main/java/uk/gov/ons/ctp/response/casesvc/service/impl/CaseServiceImpl.java
```diff
@@ -193,35 +193,33 @@ public CaseEvent createCaseEvent(final CaseEvent caseEvent, final Case newCase,
...
-        List<CaseGroup> caseGroups = caseGroupService.findCaseGroupsForExecutedCollectionExercises(targetCase);
-        processCaseCreationAndTransitionsDuringEnrolment(category, caseGroups, caseEvent, newCase);
-      }
-
-      else if (caseEvent.getCategory().equals(CategoryDTO.CategoryName.SUCCESSFUL_RESPONSE_UPLOAD)
-               || caseEvent.getCategory().equals(CategoryDTO.CategoryName.COMPLETED_BY_PHONE)) {
-        createNewCase(category, caseEvent, targetCase, newCase);
-        updateAllAssociatedBiCases(targetCase, category);
-      }
+      switch (caseEvent.getCategory()) {
```
mfcrocker: <3

PR #43, opened 2018-06-11
Changes to allow DB configuration and scaling

mfcrocker:
# Motivation and Context
It turns out that the database parameters in application-cloud.yml are overridden by CloudFoundry. This changes that!

# What has changed
`DataSourceConfiguration` has been added which forces Spring Boot to use the config rather than CloudFoundry's one from buildpacks

# How to test?
You'll need to push to CloudFoundry to see the impact of these changes. The easiest way is to set `spring.datasource.tomcat.max-idle` to a higher number than `max-active`; this will cause Spring Boot to log out a message just after the service starts that max-idle is higher than max-active and give you the figures in the .yml. If it says it's setting it to 4, it's broken (these are the defaults CF was injecting)


mfcrocker:
[Optimise DB Connection Settings](https://trello.com/c/SwXkyTRo/44-optimise-db-connection-settings)

src/main/java/uk/gov/ons/ctp/response/casesvc/CaseSvcApplication.java
```diff
@@ -1,6 +1,8 @@
 package uk.gov.ons.ctp.response.casesvc;
 
 import net.sourceforge.cobertura.CoverageIgnore;
+
+import org.apache.tomcat.jdbc.pool.DataSource;
```
tej99: Unused imports**

src/main/java/uk/gov/ons/ctp/response/casesvc/CaseSvcApplication.java
```diff
@@ -32,6 +34,7 @@
 import uk.gov.ons.ctp.common.state.StateTransitionManager;
 import uk.gov.ons.ctp.common.state.StateTransitionManagerFactory;
 import uk.gov.ons.ctp.response.casesvc.config.AppConfig;
+import uk.gov.ons.ctp.response.casesvc.config.DataSourceConfiguration;
```
tej99: **

src/main/java/uk/gov/ons/ctp/response/casesvc/CaseSvcApplication.java
tej99: Unused imports**

src/main/java/uk/gov/ons/ctp/response/casesvc/CaseSvcApplication.java
tej99: **

src/main/resources/application-cloud.yaml
```diff
@@ -53,6 +53,10 @@ spring:
       min-idle: 3
     continue-on-error: true
     schema: classpath:/schema-cloud.sql
+    user: ${cloud.services.postgresql.connection.username}
+    password: ${cloud.services.postgresql.connection.password}
+    default-schema: sample
```
gemmai95: This should be case.

src/main/java/uk/gov/ons/ctp/response/casesvc/config/DataSourceConfiguration.java
```diff
@@ -0,0 +1,40 @@
+package uk.gov.ons.ctp.response.casesvc.config;
+
+import net.sourceforge.cobertura.CoverageIgnore;
+import org.apache.tomcat.jdbc.pool.DataSource;
```
VibhorMehra: Wrong data source, should be

```
import javax.sql.DataSource;
```

src/main/resources/application-cloud.yaml
```diff
@@ -53,6 +53,10 @@ spring:
       min-idle: 3
     continue-on-error: true
     schema: classpath:/schema-cloud.sql
+    user: ${cloud.services.postgresql.connection.username}
+    password: ${cloud.services.postgresql.connection.password}
+    default-schema: case
```
VibhorMehra: Should be 

```yams
default-schema: casesvc
```

PR #46, opened 2018-06-13
Remove modifier

tej99:
# Motivation and Context

@Bean method'cloud' must not be private or final. This causes the app to fail start up.

# What has changed

Removed the final modifier.

# How to test?

# Links


mfcrocker:
[Optimise DB Connection Settings](https://trello.com/c/SwXkyTRo/44-optimise-db-connection-settings)

PR #51, opened 2018-07-02
Remove bi/generate enrolment code with case event

andrewmil:
# Motivation and Context
Currently whenever a new enrolment code is created we also create a brand new B case. This causes a number of complications and will not work/be required once we remove BI cases from the system. This PR removes the generate iac code endpoint and instead now replaces a used code attached to a case when a `GENERATE_ENROLMENT_CODE` event is posted.

# What has changed
Removed `/iac/...` POST endpoint
Added `GENERATE_ENROLMENT_CODE` category to database
Overwrite enrolment code for a case when the event is posted and the code is inactive
Added `caseiacaudit` table to database, for keeping track of old codes
Add row to caseiacaudit table whenever a new code is created

# How to test?
`mvn clean install` to run tests which should pass
Run acceptance tests with associated PR's found below
Test that generating an enrolment code with a case event works and no new cases are created

# Links
[Trello](https://trello.com/c/npXH3s9x/16-retain-current-enrolment-code-functionality-enrolment-codes)
## PR's
[casesvc-api](https://github.com/ONSdigital/rm-casesvc-api/pull/21)
[response-operations-ui](https://github.com/ONSdigital/response-operations-ui/pull/201)
[rasrm-acceptance-tests](https://github.com/ONSdigital/rasrm-acceptance-tests/pull/147)


andrewmil:
Going to hold off on merging this until data migration is better understood

PR #54, opened 2018-07-17
Added sampleUnitId to response from /cases/{uuid}

madeye-matt:
# Motivation and Context
In order that RH can retrieve sample attributes from the sample service it needs to get the sampleUnitId associated with the case.  This change includes the sampleUnitId in the response to /cases/{uuid}

# What has changed
- updated API dependency to latest to pull in new sampleUnitId in CaseDetailsDTO
- removed brittle assertion in test (counted the number of fields in JSON response)

# How to test?
- check out branch
- build locally
- make up
- make acceptance_tests
- get case id from database 
- GET http://localhost:8171/cases/{case uuid}
- ensure sampleUnitId is in response

# Links

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -149,7 +149,7 @@
         <dependency>
             <groupId>uk.gov.ons.ctp.product</groupId>
             <artifactId>casesvc-api</artifactId>
-            <version>10.49.21</version>
+            <version>10.49.22-SNAPSHOT</version>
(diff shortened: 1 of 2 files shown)


jcox-dev:
[Add sample unit id to case response](https://trello.com/c/6ZoYcFJF/192-add-sample-unit-id-to-case-response)

PR #56, opened 2018-07-19
Remove bi/always create b case at execution

gemmai95:
# Motivation and Context

This change is required to always create B cases when a collection exercise is executed. At the moment for collection exercises that are not the first only a BI case is created for business parties with someone enrolled. In order to get rid of BI cases this must always create a B case as well.

# What has changed

Add script to create B cases for case groups that do not have them (back-populate). Always create B case.

# How to test?

Tested with API changes and collection exercise when a second collection exercise is executed and a respondent is enrolled a B case is now created as well as the BI case.

# Links

[Trello] https://trello.com/c/ZWg5l0Kt

[Case-API] https://github.com/ONSdigital/rm-casesvc-api/pull/24
[Collection Exercise Service] https://github.com/ONSdigital/rm-collection-exercise-service/pull/87


src/main/java/uk/gov/ons/ctp/response/casesvc/service/impl/CaseServiceImpl.java
```diff
@@ -298,10 +297,11 @@ public void saveCaseIacAudit(final Case updatedCase) {
    * @param targetCase Case to update
    */
   private void replaceIAC(final Case targetCase) {
-    if (!internetAccessCodeSvcClientService.isIacActive(targetCase.getIac())) {
+    String iac = targetCase.getIac();
+    if (iac == null || !internetAccessCodeSvcClientService.isIacActive(iac)) {
       log.debug("Replacing existing case IAC, caseId: {}", targetCase.getId());
```
shahi645: logging level should be `info` here

src/main/java/uk/gov/ons/ctp/response/casesvc/service/impl/CaseServiceImpl.java
```diff
@@ -411,24 +411,34 @@ private void processCaseCreationAndTransitionsDuringEnrolment(
...
+      parentCase.setState(CaseState.INACTIONABLE);
+    }
+    caseRepo.saveAndFlush(parentCase);
+    createCaseCreatedEvent(parentCase, category);
+    log.info(
+        "New Case created, caseId: {}, sampleUnitType: {}",
+        parentCase.getId().toString(),
+        parentCase.getSampleUnitType().toString());
+
+    if (sampleUnitParent.getSampleUnitChildren() != null) {
```
shahi645: Don't need this second `if`?

src/main/java/uk/gov/ons/ctp/response/casesvc/service/impl/CaseServiceImpl.java
gemmai95: We tend to only use info at endpoint level. And debug for implementation etc. Or at least that is how it is in the python services.

shahi645: I think thats changing now, this seems like it should be info, but I'm not sure, so will go on your recommendation

gemmai95: I don't know we don't seem to have any guidance for the java services so I just went with what we have in Python.

src/main/java/uk/gov/ons/ctp/response/casesvc/service/impl/CaseServiceImpl.java
```diff
@@ -519,11 +527,11 @@ private void createNewCase(
   /**
    * Create the new Case.
    *
-   * @param caseData SampleUnitParent from which to create Case.
+   * @param caseData SampleUnitBase from which to create Case.
```
andrewmil: `SampleUnit`

PR #59, opened 2018-07-24
Remove bi/disable enrolment - stop creating new B case

andrewmil:
# Motivation and Context
Currently when we disable the final respondent for a business/survey we create a new B case. This creates a new enrolment code and adds a the new `B` case to the action service for future communications. A case should correspond to one response we want to receive so we should not be creating extra `B` cases on top of the existing ones for each business/collection exercise combination.

Instead we want the following behaviour in the case table when disabling the final respondent:
- No new B case is created
- Associated B case gets a new active enrolment code
- Associated B case is in `ACTIONABLE` state
- Associated B case is present in action service database (and old BI case is not)

# What has changed
Removed all logic associated with the `DISABLE_RESPONDENT_ENROLMENT` case event category. All this now does it update the status of the `BI` case to `INACTIONABLE`.
Added new category `NO_ACTIVE_ENROLMENTS` which replaces the enrolment code for the associated case and sets it's status to `ACTIONABLE` (which also removes the case from action service).

# How to test?
Run in docker with other services and associated PR's, check in database that when disabling the final respondent for a business/survey that case table is in correct state

On `mvn clean install`, should pass all unit/integration tests
Should pass acceptance tests when run alongside associated PR's

# Links
[Trello](https://trello.com/c/RlGDFpX8/18-disable-enrolment-stop-creating-b-cases-case-events-l)
## Related PR's
[casesvc-api](https://github.com/ONSdigital/rm-casesvc-api/pull/25)
[party-service](https://github.com/ONSdigital/ras-party/pull/149)

!!! This shouldn't be deployed into production until the appropriate data migration has been done to ensure that one `B` case exists for every casegroup !!!


pom.xml
```diff
@@ -150,7 +150,7 @@
         <dependency>
             <groupId>uk.gov.ons.ctp.product</groupId>
             <artifactId>casesvc-api</artifactId>
-            <version>10.49.24</version>
+            <version>10.49.25-SNAPSHOT</version>
```
andrewmil: The `SNAPSHOT` here will have to be removed once the casesvc-api PR is removed

mfcrocker:
Good to go assuming the build gets fixed!

src/test/java/uk/gov/ons/ctp/response/casesvc/service/impl/CaseCreationServiceTest.java
```diff
@@ -95,4 +96,76 @@ public void testCreateCaseAndCaseGroupWithoutChildFromMessage() {
...
+    sampleUnitParent.setSampleUnitType("B");
+    sampleUnitParent.setId(UUID.randomUUID().toString());
+
+    caseService.createInitialCase(sampleUnitParent);
+
+    // CaseGroup attributes from parent data, case from child sample unit
+    ArgumentCaptor<CaseGroup> caseGroup = ArgumentCaptor.forClass(CaseGroup.class);
+    verify(caseGroupRepo, times(1)).saveAndFlush(caseGroup.capture());
+
+    List<CaseGroup> capturedCaseGroup = caseGroup.getAllValues();
```
NickGrantONS: you could use getValue() and then you wouldn't need to have all the .get(0). bits lower down

src/test/java/uk/gov/ons/ctp/response/casesvc/service/impl/CaseCreationServiceTest.java
```diff
@@ -95,4 +96,76 @@ public void testCreateCaseAndCaseGroupWithoutChildFromMessage() {
...
+        sampleUnitParent.getSampleUnitType(), capturedCaseGroup.get(0).getSampleUnitType());
+    assertEquals(CaseGroupStatus.NOTSTARTED, capturedCaseGroup.get(0).getStatus());
+
+    ArgumentCaptor<Case> caze = ArgumentCaptor.forClass(Case.class);
+
+    verify(caseRepo, times(2)).saveAndFlush(caze.capture());
+
+    List<Case> capturedCase = caze.getAllValues();
+
+    assertEquals(UUID.class, capturedCase.get(0).getId().getClass());
```
NickGrantONS: Maybe to aid readability, you could pull out the child/parent case into a local variable so that it's clear which one you're unit testing, and which one you're ignoring (also... you could test BOTH cases if you wanted to)

src/test/java/uk/gov/ons/ctp/response/casesvc/service/impl/CaseServiceImplTest.java
```diff
@@ -331,34 +331,6 @@ public void testCreateCaseEventAgainstNonExistentCase() throws CTPException {
...
   }
 
-  /**
-   * Tries to apply an actionable event against a case already inactionable. Should allow.
-   *
-   * @throws Exception if fabricateEvent does
-   */
-  @Test
-  public void testCreateActionableEventAgainstInactionableCase() throws Exception {
-    when(caseRepo.findOne(INACTIONABLE_HOUSEHOLD_CASE_FK))
```
NickGrantONS: Why have we removed this entire unit test? It probably needs fixing, but isn't it testing a scenario which is still potentially valid?

src/main/resources/database/changes/release-10.60.0/add_no_active_enrolments_category.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO casesvc.category ("categorypk", "shortdescription", "longdescription", "eventtype", "role", "generatedactiontype", "group", "oldcasesampleunittypes", "newcasesampleunittype", "recalccollectioninstrument")
+VALUES ('NO_ACTIVE_ENROLMENTS', 'No active enrolments', 'No active enrolments remaining for case', 'ACTIVATED', NULL, NULL, NULL, 'B', NULL, NULL)
```
gemmai95: You should probably add an integration test around this.

src/test/java/uk/gov/ons/ctp/response/casesvc/service/impl/CaseServiceImplTest.java
andrewmil: The functionality this was testing was removed. If you look in `CaseServiceImpl` we've removed the `createAdHocAction` function and event though the naming isn't the best, that was what was actually being tested. 
I checked against the data in prod and that function could never be hit, I think it was something left over from census

src/test/java/uk/gov/ons/ctp/response/casesvc/service/impl/CaseCreationServiceTest.java
andrewmil: Cheers that's much better

src/test/java/uk/gov/ons/ctp/response/casesvc/service/impl/CaseCreationServiceTest.java
andrewmil: Sounds good, we probably should be testing both now so will add that 👍

src/test/java/uk/gov/ons/ctp/response/casesvc/service/impl/CaseServiceImplTest.java
NickGrantONS: ahhh... right. I guess this unit test would still hit a few of the things being tested, but I'll take your word for it that it's now redundant: I can't get my head around the naming and what exactly the test is trying to do.

src/main/resources/database/changes/release-10.60.0/add_no_active_enrolments_category.sql
andrewmil: Looking into this now 👍

andrewmil: Had a look into this. If we want to do an integration test for posting the `NO_ACTIVE_ENROLMENTS` category we're going to have to handle calls to `iac` and `action` services. We could spin these up in docker or mock them but It will probably take some time as there's no clear way of doing this in the current integration tests I don't think

andrewmil: So might leave this out for this test, a lot of this process will be changing a lot with the rest of the `BI` case work anyway

MIslam92: Umm why is it `oldcasesampleunittypes` and `newcasesampleunittype` and not either type/types for both for consistency?

src/test/java/uk/gov/ons/ctp/response/casesvc/service/impl/CaseServiceImplTest.java
```diff
@@ -1747,7 +1747,11 @@ public void testCreateInitialCaseWithSampleUnitChildren() {
     SampleUnitParent sampleUnitParent = new SampleUnitParent();
     SampleUnit sampleUnit = new SampleUnit();
     SampleUnitChildren sampleUnitChildren =
+<<<<<<< HEAD
```
gemmai95: I'm assuming you didn't mean to upload this.

PR #63, opened 2018-07-30
Remove bi/new b case migration script

andrewmil:
# Motivation and Context
Because of difficulties communicating with multiple databases in the prod/preprod environments we will now be creating a temporary table in the case service database and executing sql directly against the databases rather than leveraging a script for this.

# What has changed
Two scripts are included:
1. `SELECT` script to create temporary table required for B case creation
2. `INSERT` script to create new B case rows where appropriate

A `README.md` file is also included to explain the procedure for using these scripts

# How to test?
Set up some test data. You'll have to create some initial cases using either the acceptance tests or by manually executing collection exercises and enrolling some respondents against these. 
Since we now do create B cases you'll also have to manually delete some of the existing ones to test the scripts.

Follow the instructions in `README.md` to create the missing B cases and check that this has been successful in the database

# Links
[Trello](https://trello.com/c/qxS2f9RO/281-update-b-case-creation-script-data-migration)


migration/create-b-cases/create-b-cases.sql
```diff
@@ -0,0 +1,52 @@
...
+       SELECT DISTINCT ON (casegroupfk) casegroupfk, collectioninstrumentid
+         FROM casesvc."case"
+        WHERE sampleunittype = 'BI'
+      ) AS bicases
+     ON cg.casegrouppk = bicases.casegroupfk
+
+-- Get actionplanid from temporary table
+     INNER JOIN (
+       SELECT *
+         FROM casesvc.temp_cto
```
mfcrocker: This table should be dropped after the script is completed

migration/create-b-cases/select-temp-cto-table.sql
```diff
@@ -0,0 +1,8 @@
+SELECT ce.id AS collectionexerciseid, cto.actionplanid
```
mfcrocker: Rather than going the CSV route, you could `SELECT blahblah INTO casesvc.temp_cto` (and probably turn this into a one-script, non-manual migration in the process)

andrewmil: Afraid we can't do this. In Prod/Preprod the tables are in different databases so this can't be done in one script.

migration/create-b-cases/create-b-cases.sql
andrewmil: :+1: I'll add it into the README that the temporary table has to be dropped

migration/create-b-cases/select-temp-cto-table.sql
mfcrocker: That makes me sad :( oh well!

mfcrocker:
Happy with the convo above

PR #64, opened 2018-08-06
Find cases by samples

damien-lloyd:
# Motivation and Context
We want to find cases that relate to samples uploaded via a sample file. To do this, we search by the sample's post code and this will return cases. 

# What has changed
Endpoint for returning cases based on sample unit id

# How to test?
Run all integration tests

# Links
https://trello.com/c/uqLRhZiZ/189-sus002-find-case-by-postcode-13#


damien-lloyd:
[SUS002: Find case by Postcode (13)](https://trello.com/c/uqLRhZiZ/189-sus002-find-case-by-postcode-13)

src/main/java/uk/gov/ons/ctp/response/casesvc/domain/repository/CaseRepository.java
```diff
@@ -85,4 +87,15 @@
...
+  /**
+   * Find case by its sample unit id
+   *
+   * @param sampleUnitId
+   * @return
+   */
+  @Query(
+      value = "select caze.* from casesvc.case caze where sampleunit_id = :sampleUnitId",
+      nativeQuery = true)
+  Case findBySampleUnitId(@Param("sampleUnitId") UUID sampleUnitId);
```
benjefferies: This probably doesn't need to be a native query and we could possibly use the in clause https://docs.spring.io/spring-data/jpa/docs/1.5.0.RC1/reference/html/repository-query-keywords.html

src/main/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseEndpoint.java
```diff
@@ -103,6 +104,29 @@ public CaseEndpoint(
...
+      @RequestParam(value = "sampleUnitId") final UUID[] sampleUnitIds,
+      @RequestParam(value = "caseevents", required = false) final boolean caseevents,
+      @RequestParam(value = "iac", required = false) final boolean iac)
+      throws CTPException {
+
+    List<UUID> ids = Arrays.asList(sampleUnitIds);
+
+    List<CaseDetailsDTO> cases =
+        ids.stream()
+            .filter(e -> caseService.findCaseBySampleUnitId(e) != null)
```
benjefferies: We could possible use the in clause to do this as one query https://docs.spring.io/spring-data/jpa/docs/1.5.0.RC1/reference/html/repository-query-keywords.html

benjefferies: We should also map first so we don't have to do the query twice per ID

src/main/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseEndpoint.java
```diff
@@ -103,6 +104,29 @@ public CaseEndpoint(
     return ResponseEntity.ok(buildDetailedCaseDTO(caseObj, caseevents, iac));
   }
 
+  @RequestMapping(value = "/sampleunitids", method = RequestMethod.GET)
+  public ResponseEntity<CaseDetailsDTO[]> findCasesBySampleUnitIds(
+      @RequestParam(value = "sampleUnitId") final UUID[] sampleUnitIds,
```
atrozzo: why this one is not directly a List ?

src/main/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseEndpoint.java
```diff
@@ -103,6 +104,29 @@ public CaseEndpoint(
...
+      @RequestParam(value = "caseevents", required = false) final boolean caseevents,
+      @RequestParam(value = "iac", required = false) final boolean iac)
+      throws CTPException {
+
+    List<UUID> ids = Arrays.asList(sampleUnitIds);
+
+    List<CaseDetailsDTO> cases =
+        ids.stream()
```
atrozzo: You could have done directly Arrays.stream(array)

src/main/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseEndpoint.java
benjefferies: This makes sense, then we don't need to do anything else to stream it

src/main/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseEndpoint.java
```diff
@@ -103,6 +103,27 @@ public CaseEndpoint(
...
+  @RequestMapping(value = "/sampleunitids", method = RequestMethod.GET)
+  public ResponseEntity<List<CaseDetailsDTO>> findCasesBySampleUnitIds(
+      @RequestParam(value = "sampleUnitId") final List<UUID> sampleUnitIds,
+      @RequestParam(value = "caseevents", required = false) final boolean caseevents,
+      @RequestParam(value = "iac", required = false) final boolean iac) {
+
+    List<CaseDetailsDTO> cases =
+        sampleUnitIds
+            .stream()
+            .filter(e -> caseService.findCaseBySampleUnitId(e) != null)
```
benjefferies: If we do a map here followed by the filter we only need to do the DB lookup once

PR #65, opened 2018-08-10
Run integration tests via cobertura

benjefferies:
# Motivation and Context
This will include integration test coverage and speed the build up
because we're not running the tests twice.

# What has changed
Run IT via cobertura

# How to test?
Check codecov report

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,23 @@
 sudo: required
 
 services:
-  - docker
+- docker
 
 language: java
 jdk: openjdk8
 
-before_install: "cp .maven.settings.xml $HOME/.m2/settings.xml"
+before_install:
+- cp .maven.settings.xml $HOME/.m2/settings.xml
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,21 +1,23 @@
...
 language: java
 jdk: openjdk8
 
-before_install: "cp .maven.settings.xml $HOME/.m2/settings.xml"
+before_install:
+- cp .maven.settings.xml $HOME/.m2/settings.xml
+- mvn fmt:check
 
-script: "mvn fmt:check verify cobertura:cobertura"
+script: mvn cobertura:cobertura-integration-test
```
madeye-matt: Why remove fmt:check and verify?

benjefferies: I've moved it to the before_install, check https://github.com/ONSdigital/rm-sample-service/pull/66 for the reason :)

PR #66, opened 2018-08-13
make 'H' cases allowable for OFFLINE_RESPONSE_PROCESSED case event

davidmort:
# Motivation and Context
Currently when surveys responses/cases are receipted a case event is created of the OFFLINE_RESPONSE_PROCESSED category, only 'BI' cases are currently allowed for this. We need to make 'H' cases receipt-able within the current process by allowing OFFLINE_RESPONSE_PROCESSED case events for 'H' cases.

# What has changed
Add a DB script to add 'H' to the oldsampleunittype column for the OFFLINE_RESPONSE_PROCESSED in the category table

# How to test?
make a POST request to the receipts endpoint in rm-sdx-gateway with a caseId of a H case 

to test in the docker dev environment:
- spin up all the services
- create social data by running the setup-and-execute make target in social folder in rm-tools https://github.com/ONSdigital/rm-tools/tree/master/social-test-setup
- find an actionable 'H' case 
- make a POST request curl -X POST -d '{"caseId":"<caseId>"}' http://0.0.0.0:8191/receipts -H '<authorization details>' -H 'Content-Type:application/json'
- tail the case service logs and you should see a log line that looks like this: "Successfully created case event, casePK=2, category=OFFLINE_RESPONSE_PROCESSED, subCategory=null, createdBy=SYSTEM"

PR #67, opened 2018-08-15
Handle retries correctly

damien-lloyd:
When it comes to retries, it's very much like clockwork. A single cog that breaks can cause the entire mechanism to jigger. This was a classic case where a simple oversight in the flow caused an infinite set of retries. Initially, it was thought to be due to the spel expression being invalid or not evaluating, but what was in fact found is that at the end of each retry it was extracting the message then putting it back on the channel the retries were reading from, hence the infinite retries. 

What should have happened was at the end of each retry, the extracted message should be put onto another channel which would deposit that message into the deadletter queue. As this has now been resolved, the retries now function normally and the messages goes into the deadletter queue as expected.

How to test?

Run all the tests in RabbitSpringIntegrationDLQAndRetriesiT.java
mvn clean install
uncork a bottle.


benjefferies:
Can you add some instructions of how to test this is fixing the issue?

damien-lloyd:
[TBL 116: Identify and correct invalid retry across all Java microservices (8)](https://trello.com/c/89LJT1ie/224-tbl-116-identify-and-correct-invalid-retry-across-all-java-microservices-8)

benjefferies:
@damien-lloyd here is an example of a integration test for a queue https://github.com/ONSdigital/rm-case-service/pull/66/files#diff-6c1b5a183ab3b8eaa2f125ad8a29def4R62. This might help you write a test to simulate the retries. I am thinking that you could use WireMock to check that it hits a WireMock endpoint X times with the retry?

src/test/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseCreationReceiverImplIT.java
```diff
@@ -0,0 +1,58 @@
...
+  @MockBean private CaseService caseService;
+
+  @Test(expected = MessageTransformationException.class)
+  public void ensureFiniteRetriesOnFailedCaseNotification() throws Exception {
+    doThrow(RuntimeException.class).when(caseService).createInitialCase(any());
+
+    CaseNotification caseNotification = caseCreator.sendSampleUnit("LMS0001", "H", UUID.randomUUID());
+    // Wait for message to be DQL'd
+
+    verify(caseService, times(3)).createInitialCase(any());
```
NickGrantONS: 👍

src/test/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseCreationReceiverImplIT.java
```diff
@@ -40,18 +37,29 @@
...
+  @Autowired private MessageChannel caseTransformed;
+
   @Rule
   public WireMockRule wireMockRule =
       new WireMockRule(options().extensions(new ResponseTemplateTransformer(false)).port(18002));
 
   @MockBean private CaseService caseService;
 
-  @Test(expected = MessageTransformationException.class)
+  @Test(expected = MessageHandlingException.class)
```
NickGrantONS: I wonder if it's ever getting to the `verify`

src/test/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseCreationReceiverImplIT.java
NickGrantONS: It never _reaches_ this verify because it's expecting an exception. It'd be worth wrapping the `caseTransformed.send` in a `try...catch` so that you can then ensure the verify code is reachable. Obviously remove the `expected` exception too

src/test/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseCreationReceiverImplIT.java
```diff
@@ -0,0 +1,66 @@
...
+    sampleUnit.setId(sampleUnitId.toString());
+    sampleUnit.setActionPlanId(UUID.randomUUID().toString());
+    sampleUnit.setSampleUnitRef("LMS0004");
+    sampleUnit.setCollectionInstrumentId(UUID.randomUUID().toString());
+    sampleUnit.setPartyId(UUID.randomUUID().toString());
+    sampleUnit.setSampleUnitType("H");
+    Message<SampleUnitParent> caseMessage = new GenericMessage<>(sampleUnit);
+    caseTransformed.send(caseMessage);
+
+    verify(caseService, times(3)).createInitialCase(any());
```
NickGrantONS: I can't approve this - this line is unreachable

src/test/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseCreationReceiverImplIT.java
```diff
@@ -0,0 +1,65 @@
...
+  @Autowired private MessageChannel caseTransformed;
+
+  @Rule
+  public WireMockRule wireMockRule =
+      new WireMockRule(options().extensions(new ResponseTemplateTransformer(false)).port(18002));
+
+  @MockBean private CaseService caseService;
+
+  @Test
+  public void ensureFiniteRetriesOnFailedCaseNotification()  {
```
NickGrantONS: This test doesn't have a "then" part in the given-when-then expected structure

damien-lloyd: We can't add one since there's no concrete way of testing for retries. You just have to observe the console output and ensure the same error appears more than once.

src/main/resources/springintegration/case-creation-inbound-flow.xml
```diff
@@ -97,7 +95,7 @@
   <int:channel id="caseDlqChannelXml"/>
   <int-xml:marshalling-transformer
     input-channel="caseDlqInterimChannel" marshaller="caseCreationMarshaller"
-    output-channel="caseDlqChannel" result-transformer="caseOutboundTransformer"/>
+    output-channel="caseDlqChannelXml" result-transformer="caseOutboundTransformer"/>
```
damien-lloyd: @davidmort Here is the source of all our problems now resolved!

davidmort: 😱

PR #68, opened 2018-08-16
Find case by iac now uses caseiacaudit

madeye-matt:
Simple change to use the new caseiacaudit table to find cases by iac (instead of iac column on case table).  Note: this PR is to merge into generate_new_access_code branch not master

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/casesvc/domain/repository/CaseRepository.java
+++ b/src/main/java/uk/gov/ons/ctp/response/casesvc/domain/repository/CaseRepository.java
@@ -61,7 +61,7 @@ public interface CaseRepository extends JpaRepository<Case, Integer> {
    * @param iac the iac
    * @return the cases associated with the IAC (see above)
    */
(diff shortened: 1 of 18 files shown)


src/main/java/uk/gov/ons/ctp/response/casesvc/domain/repository/CaseRepository.java
```diff
@@ -61,7 +61,7 @@
    * @param iac the iac
    * @return the cases associated with the IAC (see above)
    */
-  List<Case> findByIac(String iac);
+  List<Case> findByIacAudits_iac(String iac);
```
benjefferies: Not sure this name change of the method helps the clarity of what it's doing

madeye-matt: LOL it's one of those JPA query methods where the it works out what to get from the method name ... it has to be called that

PR #76, opened 2018-09-03
Add script to remove multiple B cases for a case group

gemmai95:
# Motivation and Context

Add script to remove all instances of multiple B cases for case groups. It is unnecessary to have multiple B cases per case group as only the most recent one will be used. 

# What has changed

# How to test?

Add test data for multiple B cases in a case group then run script check most recent is still there.

# Links

[Trello] https://trello.com/c/NKpllwnm


andrewmil:
:+1: to the suggestion to run a separate SELECT script first to check what we're deleting ^

PR #77, opened 2018-09-04
SUS006 - Allow all IACs to be disabled for a case

liamtoozer:
# Motivation and Context

It is now possible to have multiple IACs for one case. We need to be able to disable **all** of these IACs for a disabled case event.

# What has changed

* Added a new method to disable all IACs for a case.
* Updated existing unit tests.
* Added unit tests.
* Remove two no longer valid unit tests

# How to test?

* Checkout this branch
* Run acceptance tests
* Check the acceptance test pass.

# Links

[Trello](https://trello.com/c/4ukDtQX9/216-sus006-disable-uac-when-response-status-is-changed-to-data-to-be-deleted-5)

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/casesvc/service/impl/CaseIACServiceImpl.java
+++ b/src/main/java/uk/gov/ons/ctp/response/casesvc/service/impl/CaseIACServiceImpl.java
@@ -4,7 +4,6 @@ import com.godaddy.logging.Logger;
 import com.godaddy.logging.LoggerFactory;
 import java.sql.Timestamp;
 import java.time.Clock;
-import java.util.List;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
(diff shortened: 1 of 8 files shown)


liamtoozer:
[SUS006: Disable UAC when Response Status is changed to 'Data to be deleted' (5)](https://trello.com/c/4ukDtQX9/216-sus006-disable-uac-when-response-status-is-changed-to-data-to-be-deleted-5)

src/main/java/uk/gov/ons/ctp/response/casesvc/service/impl/CaseIACServiceImpl.java
```diff
@@ -92,4 +93,14 @@ private void saveNewIacForCase(final Integer casePk, final String iac) {
             .build();
     caseIacAuditRepo.save(audit);
   }
+
+  public void disableAllIACsForCase(Case caze) {
+    caze.getIacAudits()
+        .forEach(
+            caseIacAudit -> {
+              if (!iacClient.disableIAC(caseIacAudit.getIac())) {
+                log.with("caseId", caze.getId()).error("Failed to disable an IAC for case");
```
damien-lloyd: Throw an exception here

AdamHawtin: If one IAC fails to disable we still want to attempt to disable any other remaining. There's not really anything else this service can do at this point other than log the failure.

AdamHawtin: We resolved this as the IAC's are retrieved from a `NOT NULL` database column they are already null checked.

PR #82, opened 2018-09-18
Add data migration to remove completed cases from action

gemmai95:
# Motivation and Context

This data migration can be used to remove cases that are complete but have not been removed from the action service. This should stop people getting reminders they shouldn't when another respondent for that business has responded.

# What has changed

# How to test?

# Links


andrewmil:
Since the actual table updates we're doing are in action I think the scripts should sit there rather than in case

PR #87, opened 2018-09-26
Update snapshot version to 10.49.22

LJBabbage:
We recently had to do a bug fix off an earlier release and create a jar from it in artifactory. Because of that we are out of step and the snapshot version needs updating


jcox-dev:
[SUS013 Make a Collection Exercise unavailable after the "End Date" is reached (5)](https://trello.com/c/sC7bhXJn/213-sus013-make-a-collection-exercise-unavailable-after-the-end-date-is-reached-5)

PR #91, opened 2018-10-12
Updated case events by case id endpoint to take optional category

shahi645:
# Motivation and Context
Rops users want to be able to get the timestamp for a completed `seft` or `eq`. We have decided for the first iteration to only return the first completed timestamp for SEFT. 

# What has changed
Updated endpoint to take optional category and return case events based on category, along with said created date stamp. 

# How to test?
Run in docker dev and hit endpoint to see if matched case events to categories and case id's are returned. 

Wanted categories:
`OFFLINE_RESPONSE_PROCESSED`
`COMPLETED_BY_PHONE`
`SUCCESSFUL_RESPONSE_UPLOAD`

Note this end point will return all categories sent if in DB. 

# Links
[TRELLO CARD](https://trello.com/c/sNa7NP3W/462-create-endpoint-in-case-service-to-expose-completion-details-incl-timestamp-m)


src/main/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseEndpoint.java
```diff
@@ -284,15 +284,18 @@ private void createNewEventForAccessCodeAuthAttempt(final Case targetCase) throw
...
-   * @return the case events found
+   * @param category type to collect case event timestamp
+   * @return the case events found, with optional timestamps (category required for this)
    * @throws CTPException something went wrong
    */
   @RequestMapping(value = "/{caseId}/events", method = RequestMethod.GET)
   public ResponseEntity<List<CaseEventDTO>> findCaseEventsByCaseId(
-      @PathVariable("caseId") final UUID caseId) throws CTPException {
+      @PathVariable("caseId") final UUID caseId,
+      @RequestParam(value = "category", required = false) final List<String> categories)
```
NickGrantONS: actually, I was being dumb... we can use the enumeration type and Spring will make sure that only valid enum values are chosen and return a 400 bad request if garbage is sent

gemmai95:
Do you not want No Longer Required as well?

src/main/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseEndpoint.java
gemmai95: I still think this should be category name rather than string. Will we want a list or just one category?

shahi645: So, tried that, it throws a 500 instead of a 400 when the category name doesnt exist, which is wrong behaviour. To get around that needs to be passed as a string so just keeping it this way for now

shahi645:
@gemmai95 No, although this endpoint can be used to get that info, but i think we only want completions

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
```diff
@@ -599,4 +605,16 @@ private CaseGroup createNewCaseGroup(SampleUnitParent caseGroupData) {
...
   }
+
+  public List<CaseEvent> findCaseEventsByCaseFKAndCategory(Integer casePK, List<String> categories)
+      throws CTPException {
+    try {
+      Set<CategoryName> categoryNames =
+          categories.stream().map(CategoryDTO.CategoryName::fromValue).collect(Collectors.toSet());
+      return caseEventRepo.findByCaseFKAndCategoryInOrderByCreatedDateTimeDesc(
+          casePK, categoryNames);
+    } catch (IllegalArgumentException exc) {
```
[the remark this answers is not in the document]
MIslam92: Do we still need this @shahi645 ?

src/test/java/uk/gov/ons/ctp/response/casesvc/service/CaseServiceTest.java
```diff
@@ -1071,6 +1079,31 @@ public void testRespondentEnrolledCaseEventWithNoActionPlansThrowsException()
...
+
+    // Then
+    verify(caseEventRepo)
+        .findByCaseFKAndCategoryInOrderByCreatedDateTimeDesc(eq(123), argumentCaptor.capture());
+    assertTrue(argumentCaptor.getValue().contains(EQ_LAUNCH));
+    assertThat(argumentCaptor.getValue().size()).isEqualTo(1);
+  }
+
+  @Test(expected = CTPException.class)
+  public void testFindByCaseFKAndCategoryBlowsUpWhenStringDoesNotMapToEnumValue()
```
MIslam92: Are you not asserting this test against something?

API.md
```diff
@@ -269,6 +269,12 @@ An `HTTP 404 Not Found` status code is returned if the case with the specified u
 ## List Case Events for Case
 * `GET /cases/{caseId}/events` will return a list of case events for the case with given case ID.
 
+&mdash;an optional `category` string query parameter can be used to specify that the JSON response include matching events with the case for said case id. For example:
+
+* `GET /cases/{caseId}/events?category=OFFLINE_RESPONSE_PROCESSED`
+
+If this query parameter is omitted all events for said case ID will be returned.
+
```
MIslam92: "In this scenario all events for a given case ID can be retrieved by using the `GET /cases/{caseId}/events` endpoint"

Might want to add this scenario example like how we do it for the rest of the `API.md`, wording is up to you.

shahi645: I think its pretty explicit tbh

MIslam92: It is just how its done throughout the rest of the file etc.

PR #98, opened 2018-11-02
Fix for intermittent test failure

ryangrundy7:
# Motivation and Context
Theres an intermittent bug in the case service tests where it gets stuck attempting to contact the IAC service wiremock. This only happened occasionally so its hard to test.

# What has changed
- Added classrule to the wiremock
- Added a thread sleep to allow the wiremock to shutdown and wake up between tests.

# How to test?
- running `mvn clean install` and it passing. When testing this I had to run it constantly just in case it happened again, so you may need to keep running the tests just to be thorough. Only merge when confident you can't replicate the bug.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -48,7 +48,7 @@ services:
       test: ["CMD", "curl", "-f", "http://localhost:8145/info"]
       interval: 30s
       timeout: 10s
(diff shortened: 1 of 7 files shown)


NickGrantONS:
Hung on the second time too

PR #101, opened 2018-11-23
Bump API version

NickGrantONS:
# Motivation and Context
EQ will start sending us messages containing a party ID at some point in future, which could cause the messages to be rejected, because we're not expecting the extra element in the Rabbit message.

This PR prepares the Case service with the new version of the Case API, containing the new optional party ID element in the XSD/XML.

# What has changed
Bumped the version of the API we're using, which contains the new XML schema.

# How to test?
Run the ATs to check for no regression.

# Links
Trello: https://trello.com/c/gV6X31s2/590-prepare-case-service-for-casereceipt-messages-from-eq-containing-party-id

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -158,7 +158,7 @@
         <dependency>
             <groupId>uk.gov.ons.ctp.product</groupId>
             <artifactId>casesvc-api</artifactId>
-            <version>10.49.30</version>
+            <version>10.49.32</version>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -158,7 +158,7 @@
         <dependency>
             <groupId>uk.gov.ons.ctp.product</groupId>
             <artifactId>casesvc-api</artifactId>
-            <version>10.49.30</version>
+            <version>10.49.32</version>
```
necrophonic: I'm probably missing context (or coffee) but I assume there's a reason it's skipped `.31`?

NickGrantONS: `.31` probably contained stuff we ended up not needing. We're also not very good at keeping our API versions up to date.

PR #102, opened 2019-01-11
Updating receipt receiver to use user Id

ryangrundy7:
# Motivation and Context
To allow the userid to get passed from rm-sdx gateway to case service, we need to update the case receipt receiver to use the case receipt party id instead of the existing case party id.
# What has changed
- Changed partyid from the existing case to the case receipt

# How to test?
Need to run with [RM-SDX Gateway PR](https://github.com/ONSdigital/rm-sdx-gateway/pull/21)

To test this locally I send a POST request to RM-SDX Gateway at the http://localhost:8191/receipts which includes the caseid and partyid of the respondent which you can get if you debug frontstage when you click access survey for an eQ. You need to first make sure an eQ has been set to in progress before you can do this. This can be done by clicking access survey. If you look in the case events table then the partyId should be a respondent and looking at the response status for that eQ should show the respondent.
# Links
[trello](https://trello.com/c/Hd4vgVp5)

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/casesvc/message/CaseReceiptReceiver.java
+++ b/src/main/java/uk/gov/ons/ctp/response/casesvc/message/CaseReceiptReceiver.java
@@ -72,7 +72,7 @@ public class CaseReceiptReceiver {
     } else {
       CaseEvent caseEvent = new CaseEvent();
       String partyId = caseReceipt.getPartyId();
-      if (partyId != null) {
+      if (partyId != null && !partyId.isEmpty()) {
         Map<String, String> metadata = new HashMap<>();
         metadata.put("partyId", partyId);
         caseEvent.setMetadata(metadata);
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/ons/ctp/response/casesvc/message/CaseReceiptReceiver.java
```diff
@@ -71,10 +71,10 @@ public void process(CaseReceipt caseReceipt) throws CTPException {
       log.with("case_id", caseId).error(EXISTING_CASE_NOT_FOUND);
     } else {
       CaseEvent caseEvent = new CaseEvent();
-      UUID partyId = existingCase.getPartyId();
+      String partyId = caseReceipt.getPartyId();
       if (partyId != null) {
```
NickGrantONS:
```suggestion
      if (partyId != null && !partyId.isEmpty()) {
```

PR #105, opened 2019-02-07
Performance improvements

ryangrundy7:
# Motivation and Context
In an attempt to make the to-do page on frontstage even quicker, We've decided that adding an index  on the case table for the partyid would help increase the speed. It will also now pass in a IAC boolean so it can skip adding the iac which removes another DB calls
# What has changed
- Added an index on the case table for the partyId column
- Added boolean check on IAC so it can be skipped

# How to test?
- Run with [frontstage PR](https://github.com/ONSdigital/ras-frontstage/pull/429)
- Run the acceptance tests
- Open up the todo page and make sure everything is working such as accessing survey and completing it.

# Links
[Trello](https://trello.com/c/5xsPP92s)

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
+++ b/src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
@@ -156,10 +156,13 @@ public class CaseService {
    * @param partyId the partyId
    * @return the cases for the partyId
    */
-  public List<Case> findCasesByPartyId(final UUID partyId) {
+  public List<Case> findCasesByPartyId(final UUID partyId, boolean iac) {
     log.debug("Entering findCasesByPartyId");
     List<Case> cazes = caseRepo.findByPartyId(partyId);
-    cazes.stream().forEach(c -> c.setIac(caseIacAuditService.findCaseIacByCasePK(c.getCasePK())));
+
+    if (iac) {
+      cazes.stream().forEach(c -> c.setIac(caseIacAuditService.findCaseIacByCasePK(c.getCasePK())));
+    }
 
     return cazes;
   }
--- a/src/main/resources/database/changes/release-19/changelog.yml
+++ b/src/main/resources/database/changes/release-19/changelog.yml
@@ -6,6 +6,6 @@ databaseChangeLog:
     changes:
     - sqlFile:
         comment: Adding metadata to case events table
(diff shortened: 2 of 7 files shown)


insacuri:
The code looks reasonable, though I'm not well versed enough in it to approve it I feel.
Just to make sure I understand, when we search for cases, we already know if there's an iac for it already?  Which is why we can only do that extra step if we know that one needs to be set?

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
```diff
@@ -156,10 +156,13 @@ public Case findCaseByIac(final String iac) throws CTPException {
    * @param partyId the partyId
    * @return the cases for the partyId
    */
-  public List<Case> findCasesByPartyId(final UUID partyId) {
+  public List<Case> findCasesByPartyId(final UUID partyId, boolean iac) {
     log.debug("Entering findCasesByPartyId");
     List<Case> cazes = caseRepo.findByPartyId(partyId);
```
owencjones: Is this named `cazes` to avoid clash with another `cases` variable in scope?

src/main/resources/database/changes/release-19/changelog.yml
```diff
@@ -6,6 +6,6 @@ databaseChangeLog:
     changes:
     - sqlFile:
         comment: Adding metadata to case events table
-        path: add-metadata-to-caseevents.sql
+        path:  add-metadata-to-caseevents.sql
```
owencjones: No need for this extra space?

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
NickGrantONS: no, it's named `cazes` because when we have an object of class `Case` we can't name it `case` because `case` is a reserved word... the convention for the name of a collection of `Case` objects has thus become `cazes` which is perhaps something which could be addressed in future, but isn't really in the scope of this pull request, although I appreciate the gravity of the situation of having several collections named `cazes` and the urgent need to change it everywhere it occurs.

src/main/resources/database/changes/release-19/changelog.yml
NickGrantONS: No need at all. Completely superfluous.

PR #108, opened 2019-02-19
589 performance improvements to case service

ryangrundy7:
# Motivation and Context
In the Case Service, it makes repeated calls to the collection exercise service of the same collex. Instead of wasting time repeating calls, we can make the method cacheable and instead it can use the cached version of the collection exercise if it's already there.

The other part of this ticket was to up the batch size of the case distributor and increase the messaging consumerThreads. Until we can use a performance environment, it's best to leave this bit and just add the cacheable to the Case service.

# What has changed
- Added `Cacheable` to `getCollectionExercise`
- Moved `application-test.yml` into the testing folder.

# How to test?
- Execute a large collection exercise and check how many times the `Retrieving collection exercise` log line gets printed.

# Links
[Trello](https://trello.com/c/2xfURaiX)


src/main/java/uk/gov/ons/ctp/response/casesvc/client/CollectionExerciseSvcClient.java
```diff
@@ -44,6 +45,7 @@
    * @param collectionExerciseId the UUID to search by
    * @return the asscoaited CollectionExercise
    */
+  @Cacheable("collectionexercises")
```
mfcrocker: Is it worth making `COLLEX_CACHE_NAME` public in `CaseSvcApplication` so that, rather than using a string literal here, we could instead just call `CaseSvcApplication.COLLEX_CACHE_NAME` to avoid errors if we ever change this name?

src/main/java/uk/gov/ons/ctp/response/casesvc/CaseSvcApplication.java
```diff
@@ -247,4 +255,17 @@ public DistributedLockManager reportDistributedLockManager(RedissonClient rediss
...
+  @Bean
+  public CacheManager cacheManager() {
+    return new ConcurrentMapCacheManager(COLLEX_CACHE_NAME);
+  }
+
+  @CacheEvict(
+      allEntries = true,
+      cacheNames = {COLLEX_CACHE_NAME})
+  @Scheduled(fixedDelay = 60000)
+  public void cacheEvict() {}
```
NickGrantONS: Maybe put a comment in the method to say "this is getting rid of cache entries, in case anything's been changed - we imagine that the maximum of 1 minute delay to seeing changes reflected in the case service will not cause any issues"

src/main/java/uk/gov/ons/ctp/response/casesvc/client/CollectionExerciseSvcClient.java
NickGrantONS: 👍

ryangrundy7: @mfcrocker Done

src/main/java/uk/gov/ons/ctp/response/casesvc/CaseSvcApplication.java
ryangrundy7: @NickGrantONS Done

NickGrantONS:
Why did this branch get deleted?

ryangrundy7:
@NickGrantONS we had problems with travis so deleted this pr to fix it, this [PR](https://github.com/ONSdigital/rm-case-service/pull/109) was the replacement

PR #114, opened 2019-09-05
Added ability to limit cases by survey, not returning older ones

AndrewTorrance:
# Motivation and Context
Response ops UI was getting all cases for all surveys for a specific ru , then getting all the details for all the cases , then only showing 12 per survey in the response ops view. This ticket was to restrict the returned cases to be the last n cases per survey. So that the redundant  look ups in the UI where not needed . 

# What has changed
Added an extra parameter on the call. If the parameter is missing the code follows the old route. If t is present then it orders the get case by party by create date descending and then only returns the requested number per survey . So older cases are not returned. Currently the UI sets this to 12 so it can show 12 cases per survey per ru , which is ok for the last year of monthly surveys. (The branch name says pagination as this was the first approach but that did it across all surveys)
 

Most of the case selection and ordering is via JPA . JPQL could not cope with the complex query needed to implement this in SQL , so a function was added to limit the result set in code as a pragmatic compromise. This was subject to a new unit test

# How to test?
Run acceptance tests to get data in a good state.
look at the collection exercises for an ru with multiple collexes. Stop response ops . Either export MAX_CASES_RETRIEVED_PER_SURVEY = 1 and rerun and verify that only the last collex is shown , or run it in a debugger and manually set it to 1. Verify that the older collexes are the ones dropped,

# Links
https://trello.com/c/IRgw9xaj

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseEndpoint.java
+++ b/src/main/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseEndpoint.java
@@ -132,16 +132,28 @@ public final class CaseEndpoint implements CTPEndpoint {
    * @param partyId to find by
    * @param caseevents flag used to return or not CaseEvents
    * @param iac flag used to return or not the iac
+   * @param max_cases_per_survey the maximum number of collection exercises to return per survey
    * @return the cases found
    */
-  @Deprecated // See findCases(sampleUnitId, partyId)
   @RequestMapping(value = "/partyid/{partyId}", method = RequestMethod.GET)
   public ResponseEntity<List<CaseDetailsDTO>> findCasesByPartyId(
       @PathVariable("partyId") final UUID partyId,
       @RequestParam(value = "caseevents", required = false) final boolean caseevents,
-      @RequestParam(value = "iac", required = false) final boolean iac) {
-    log.with("party_id", partyId).debug("Retrieving cases by party");
-    List<Case> casesList = caseService.findCasesByPartyId(partyId, iac);
+      @RequestParam(value = "iac", required = false) final boolean iac,
+      @RequestParam(value = "max_cases_per_survey", required = false)
+          final Integer maxCasesPerSurvey) {
+
+    List<Case> casesList;
+    if (maxCasesPerSurvey != null) {
+      log.with("party_id", partyId)
+          .with("max_cases_per_survey", maxCasesPerSurvey)
+          .info("Retrieving cases by party");
+      casesList = caseService.findCasesByPartyIdLimitedPerSurvey(partyId, iac, maxCasesPerSurvey);
+
+    } else {
+      log.with("party_id", partyId).info("Retrieving cases by party");
+      casesList = caseService.findCasesByPartyId(partyId, iac);
+    }
 
     if (CollectionUtils.isEmpty(casesList)) {
       return ResponseEntity.noContent().build();
--- a/src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
+++ b/src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
@@ -3,7 +3,9 @@ package uk.gov.ons.ctp.response.casesvc.service;
 import com.godaddy.logging.Logger;
 import com.godaddy.logging.LoggerFactory;
 import java.sql.Timestamp;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
@@ -157,7 +159,7 @@ public class CaseService {
    * @return the cases for the partyId
    */
   public List<Case> findCasesByPartyId(final UUID partyId, boolean iac) {
-    log.debug("Entering findCasesByPartyId");
+    log.info("Entering findCasesByPartyId");
     List<Case> cazes = caseRepo.findByPartyId(partyId);
 
     if (iac) {
@@ -165,6 +167,86 @@ public class CaseService {
     }
 
     return cazes;
+  }
+
+  /**
+   * Find the paginated cases for a partyId ordered by create date descending.
+   *
+   * @param partyId the partyId
+   * @param maxCasesPerSurvey the maximum number of cases per survey to return
+   * @return the cases for the partyId
+   */
+  public List<Case> findCasesByPartyIdLimitedPerSurvey(
+      final UUID partyId, boolean iac, int maxCasesPerSurvey) {
+    log.info("Entering findCasesByPartyIdLimitedPerSurvey");
+
+    List<Case> cazes =
+        limitCasesPerSurvey(
+            caseRepo.findByPartyIdOrderByCreatedDateTimeDesc(partyId), maxCasesPerSurvey);
+
+    if (iac) {
+      cazes.stream().forEach(c -> c.setIac(caseIacAuditService.findCaseIacByCasePK(c.getCasePK())));
+    }
+
+    return cazes;
+  }
+
+  /**
+   * Return only the first n instances of case per survey The query to this in the JPA is too
+   * complex and uses SQL constructs that JPQL does not support like Row Number and Over. Hence in
+   * code
+   *
+   * @param raw, the list of cases to be potentially limited
+   * @param maxCasesPerSurvey the maximum number of cases per survey to return
+   * @return the cases for the partyId
+   */
+  public List<Case> limitCasesPerSurvey(List<Case> raw, int maxCasesPerSurvey) {
+
+    log.with("maxCasesPerSurvey", maxCasesPerSurvey).info("Entering limitCasesPerSurvey");
+
+    List<Case> limitedResults = new ArrayList();
+    HashMap<UUID, Integer> surveyCounts = new HashMap<>();
+
+    raw.stream()
+        .forEach(
+            c -> tryAddCaseToLimitedResults(c, limitedResults, surveyCounts, maxCasesPerSurvey));
+
+    return limitedResults;
+  }
+
+  /**
+   * Adds a case to a result set if the number per survey is less than maximum
+   *
+   * @param current , the current Case to consider adding
+   * @param limitedResults The limited number of results
+   * @param surveyCounts A hashmap of survey ids and counts , used to keep a running count
+   * @param maxCasesPerSurvey the maximum number of cases per survey to return
+   * @return void , adds results to passed in limited as it goes
+   */
+  private void tryAddCaseToLimitedResults(
+      Case current,
+      List<Case> limitedResults,
+      HashMap<UUID, Integer> surveyCounts,
+      int maxCasesPerSurvey) {
+
+    CaseGroup currentGroup = caseGroupRepo.findById(current.getCaseGroupId());
+    UUID currentSurveyId = currentGroup.getSurveyId();
+
+    log.with("currentSurveyId", currentSurveyId).info("Entering tryAddCaseToLimitedResults");
+
+    if (!surveyCounts.containsKey(currentSurveyId)) {
+      surveyCounts.put(currentSurveyId, 0);
+    }
+    int currentCount = surveyCounts.get(currentSurveyId);
+
+    if (currentCount < maxCasesPerSurvey) {
+      surveyCounts.replace(currentSurveyId, ++currentCount);
+      limitedResults.add(current);
+    } else {
+      log.with("surveyId", currentSurveyId.toString())
+          .with("count", currentCount)
+          .info("Survey rejected");
+    }
   }
 
   /**
--- a/src/test/java/uk/gov/ons/ctp/response/casesvc/service/CaseServiceTest.java
+++ b/src/test/java/uk/gov/ons/ctp/response/casesvc/service/CaseServiceTest.java
@@ -714,6 +714,60 @@ public class CaseServiceTest {
   }
 
   /**
+   * Tests that with multiple survey ids then the limit per survey matches maxCasesPerSurvey and
+   * that the later cases in the list are rejected This suffices because SpringBoot implements the
+   * findByDateTime desc , so we only need validate the later ones are removed
+   *
+   * @throws Exception
+   */
+  @Test
+  public void testLimitCasesPerSurveyRemovesCasesFromEndOfList() throws Exception {
+
+    int maxCasesPerSurvey = 2;
+
+    // Arrange: Create a List<Case> of 6 entries corresponding to two groups
+
+    UUID groupUUIDOne = UUID.randomUUID();
+    UUID groupUUIDTwo = UUID.randomUUID();
+
+    Case case_one = makeCase(groupUUIDOne);
+    Case case_two = makeCase(groupUUIDOne);
+    Case case_three = makeCase(groupUUIDOne);
+    Case case_four = makeCase(groupUUIDTwo);
+    Case case_five = makeCase(groupUUIDTwo);
+    Case case_six = makeCase(groupUUIDTwo);
+
+    List<Case> caseList = new ArrayList<>();
+
+    caseList.add(case_one);
+    caseList.add(case_two);
+    caseList.add(case_three);
+    caseList.add(case_four);
+    caseList.add(case_five);
+    caseList.add(case_six);
+
+    CaseGroup caseGroupOne = makeCaseGroupWithSurveyId(UUID.randomUUID());
+    CaseGroup caseGroupTwo = makeCaseGroupWithSurveyId(UUID.randomUUID());
+
+    when(caseGroupRepo.findById(groupUUIDOne)).thenReturn(caseGroupOne);
+    when(caseGroupRepo.findById(groupUUIDTwo)).thenReturn(caseGroupTwo);
+
+    // Act: limit those to 2 cases per survey id
+
+    List<Case> results = caseService.limitCasesPerSurvey(caseList, maxCasesPerSurvey);
+
+    // Assert: First the right count
+    assertEquals(results.size(), 4);
+
+    // Assert: Secondly that the Cases at the start of the list are present
+    assertEquals(results.get(0).getId(), case_one.getId());
+    assertEquals(results.get(1).getId(), case_two.getId());
+
+    assertEquals(results.get(2).getId(), case_four.getId());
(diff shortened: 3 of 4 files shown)


API.md
```diff
@@ -23,9 +23,13 @@ For the endpoints that return the details of a case:
 
 &mdash;an optional `iac` boolean query parameter can be used to specify that the JSON response include the unique access code associated with the case. For example:
 
-* `GET /cases/partyid/{partyId}?iac=true`
+* `GET /cases/partyid/{partyId}?iac=true?max_cases_per_survey=<n>`
```
mfcrocker: I would refactor this to talk about each optional parameter separately to better fit the sentence above. I'd also like to see the parameterised bit written as `{n}` rather than `<n>` for consistency

API.md
```diff
@@ -23,9 +23,13 @@ For the endpoints that return the details of a case:
...
 
-If this query parameter is omitted the unique access code **will not** be returned with the case details. The JSON examples provided for the three endpoints mentioned above include the code for illustration purposes.
+If the iac query parameter is omitted the unique access code **will not** be returned with the case details. The JSON examples provided for the three endpoints mentioned above include the code for illustration purposes.
+
+The max_cases_per_survey is an optional parameter. If not present then all cases for that party will be returned
+If present as an integer then only the most recent <n> cases per survey will
```
mfcrocker: `{n}` as above and the newline here is weird

src/main/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseEndpoint.java
```diff
@@ -132,16 +132,28 @@ public CaseEndpoint(
    * @param partyId to find by
    * @param caseevents flag used to return or not CaseEvents
    * @param iac flag used to return or not the iac
+   * @param max_cases_per_survey the maximum number of collection exercises to return per survey
    * @return the cases found
    */
-  @Deprecated // See findCases(sampleUnitId, partyId)
```
mfcrocker: Why are we un-deprecating this method? Surely we should be modifying `findCases(sampleUnitId, partyId)` instead?

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
```diff
@@ -157,7 +159,7 @@ public Case findCaseByIac(final String iac) throws CTPException {
    * @return the cases for the partyId
    */
   public List<Case> findCasesByPartyId(final UUID partyId, boolean iac) {
-    log.debug("Entering findCasesByPartyId");
+    log.info("Entering findCasesByPartyId");
```
mfcrocker: This log line is just noise, I would just delete it

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
```diff
@@ -167,6 +169,86 @@ public Case findCaseByIac(final String iac) throws CTPException {
...
+  /**
+   * Find the paginated cases for a partyId ordered by create date descending.
+   *
+   * @param partyId the partyId
+   * @param maxCasesPerSurvey the maximum number of cases per survey to return
+   * @return the cases for the partyId
+   */
+  public List<Case> findCasesByPartyIdLimitedPerSurvey(
+      final UUID partyId, boolean iac, int maxCasesPerSurvey) {
+    log.info("Entering findCasesByPartyIdLimitedPerSurvey");
```
mfcrocker: As above

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
```diff
@@ -167,6 +169,86 @@ public Case findCaseByIac(final String iac) throws CTPException {
...
+
+    List<Case> cazes =
```
mfcrocker: 🎨 I would just call this `cases` - we tend to only use these weird spellings when it conflicts with a keyword (like `clazz`)

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
```diff
@@ -167,6 +169,86 @@ public Case findCaseByIac(final String iac) throws CTPException {
...
+  /**
+   * Return only the first n instances of case per survey The query to this in the JPA is too
+   * complex and uses SQL constructs that JPQL does not support like Row Number and Over. Hence in
+   * code
+   *
+   * @param raw, the list of cases to be potentially limited
+   * @param maxCasesPerSurvey the maximum number of cases per survey to return
+   * @return the cases for the partyId
+   */
+  public List<Case> limitCasesPerSurvey(List<Case> raw, int maxCasesPerSurvey) {
```
mfcrocker: Is this method going to be used outside of this class? I would personally make this private

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
```diff
@@ -167,6 +169,86 @@ public Case findCaseByIac(final String iac) throws CTPException {
...
+
+    log.with("maxCasesPerSurvey", maxCasesPerSurvey).info("Entering limitCasesPerSurvey");
```
mfcrocker: Noise

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
```diff
@@ -167,6 +169,86 @@ public Case findCaseByIac(final String iac) throws CTPException {
...
+
+    List<Case> limitedResults = new ArrayList();
+    HashMap<UUID, Integer> surveyCounts = new HashMap<>();
+
+    raw.stream()
```
mfcrocker: I'm not convinced we need a new method for `tryAddCaseToLimitedResults` - this method is very small and this is the only method that will call TACTLR. I'd just write this as `for (Case case : raw) { ... logic ... }` rather than getting cute with streaming it.

I think the case for this is shown by the fact we have to keep passing in `limitedResults` and `surveyCounts` - it's just not a natural candidate for extraction to a method.

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
```diff
@@ -167,6 +169,86 @@ public Case findCaseByIac(final String iac) throws CTPException {
...
+   * @param maxCasesPerSurvey the maximum number of cases per survey to return
+   * @return void , adds results to passed in limited as it goes
+   */
+  private void tryAddCaseToLimitedResults(
+      Case current,
+      List<Case> limitedResults,
+      HashMap<UUID, Integer> surveyCounts,
+      int maxCasesPerSurvey) {
+
+    CaseGroup currentGroup = caseGroupRepo.findById(current.getCaseGroupId());
```
mfcrocker: This is a very expensive transaction if we're doing it inside the `raw.forEach` loop. Is there a way we can find this information out once rather than going back to the database every time?

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
```diff
@@ -167,6 +169,86 @@ public Case findCaseByIac(final String iac) throws CTPException {
...
+    UUID currentSurveyId = currentGroup.getSurveyId();
+
+    log.with("currentSurveyId", currentSurveyId).info("Entering tryAddCaseToLimitedResults");
```
mfcrocker: Noise (and also not at the point where we enter `tryAddCaseToLimitedResults`)

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
```diff
@@ -167,6 +169,86 @@ public Case findCaseByIac(final String iac) throws CTPException {
...
+    if (!surveyCounts.containsKey(currentSurveyId)) {
+      surveyCounts.put(currentSurveyId, 0);
+    }
+    int currentCount = surveyCounts.get(currentSurveyId);
+
+    if (currentCount < maxCasesPerSurvey) {
+      surveyCounts.replace(currentSurveyId, ++currentCount);
+      limitedResults.add(current);
+    } else {
+      log.with("surveyId", currentSurveyId.toString())
```
mfcrocker: Is this important to log? Personally I don't think so, but if it is, I'd go with a better description than `Survey rejected` - this makes me think something properly bad has happened when in fact we're just rate limiting the results

src/test/java/uk/gov/ons/ctp/response/casesvc/service/CaseServiceTest.java
```diff
@@ -713,6 +713,60 @@ public void testEventRespondentEnrolled() throws Exception {
...
+
+    CaseGroup caseGroupOne = makeCaseGroupWithSurveyId(UUID.randomUUID());
+    CaseGroup caseGroupTwo = makeCaseGroupWithSurveyId(UUID.randomUUID());
+
+    when(caseGroupRepo.findById(groupUUIDOne)).thenReturn(caseGroupOne);
+    when(caseGroupRepo.findById(groupUUIDTwo)).thenReturn(caseGroupTwo);
+
+    // Act: limit those to 2 cases per survey id
+
+    List<Case> results = caseService.limitCasesPerSurvey(caseList, maxCasesPerSurvey);
```
mfcrocker: As far as I can tell this is the only method that uses the public nature of `limitCasesPerSurvey` - the usual way of testing this as a private method would be to mock out `caseRepo.findByPartyIdOrderByCreatedDateTimeDesc` like you have just above this and then calling the public method `findCasesByPartyIdLimitedPerSurvey`, thereby testing the whole lot rather than just the filtering method.

API.md
AndrewTorrance: Good catch , changing

API.md
AndrewTorrance: changing

src/main/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseEndpoint.java
AndrewTorrance: In the commit that added it Ben mentions it was only use was in tests , and no mention of why he did not use in in actual code .  Given this was over a year ago and the existing endpoint was in use It seemed a little outside of the scope this task to implement his deprecation ?

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
AndrewTorrance: Will turn them down to debug , but it is following the existing pattern of logging on entry

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
AndrewTorrance: as above

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
insacuri: It's noise unless you put some context with it, if you log it with the party id then it'd be useful

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
insacuri: As above (add context and it'll be useful!)

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
insacuri: yes please!  Java using weird spellings for words was never my favourite practice >_<

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
AndrewTorrance: Yep , bothers me too. Not sure about making it cacheable at the repo though , so that leaves a manual cache I think ?

src/test/java/uk/gov/ons/ctp/response/casesvc/service/CaseServiceTest.java
AndrewTorrance: Its one of those odd ones in that I wanted a unit test (possibly need a higher level one as well)  But couldn't as it was private . You could use this function to modify any list of cases per survey so it could be a useful utility. Protected is possibly a better match.

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
AndrewTorrance: As a general point , I basically followed the existing convention on these log lines on entry as its my first foray into case. Ok I upped some from debug to info( and am now turning them back ). I think its worth agreeing what we want and applying it across the board in the service?

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
AndrewTorrance: Changing

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
AndrewTorrance: changed

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
insacuri: :art: raw isn't a great name for this variable, maybe unrefinedCaseList?  Just something that's actually obvious at a glance.  I only knew what it was when I looked at the docstring

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
mfcrocker: I would never consider this useful information in production at debug or info - it's the sort of thing I add to my code while I'm writing it and then get rid of before checking in.

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
mfcrocker: As above!

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
mfcrocker: Still needs looking at

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
```diff
@@ -167,6 +169,78 @@ public Case findCaseByIac(final String iac) throws CTPException {
...
+  /**
+   * Return only the first n instances of case per survey The query to do this in the JPA is too
+   * complex and uses SQL constructs that JPQL does not support like Row Number and Over. Hence in
+   * code
+   *
+   * @param raw, the list of cases to be potentially limited
+   * @param maxCasesPerSurvey the maximum number of cases per survey to return
+   * @return the cases for the partyId
+   */
+  private List<Case> limitCasesPerSurvey(List<Case> raw, int maxCasesPerSurvey) {
```
mfcrocker: `raw` still not a good variable name as per @insacuri's comment

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
```diff
@@ -167,6 +169,78 @@ public Case findCaseByIac(final String iac) throws CTPException {
...
+
+    log.with("maxCasesPerSurvey", maxCasesPerSurvey).debug("Entering limitCasesPerSurvey");
```
mfcrocker: This log line is still no use to us

src/main/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseEndpoint.java
mfcrocker: Sure, that's fine with me

src/main/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseEndpoint.java
```diff
@@ -132,16 +132,28 @@ public CaseEndpoint(
...
-      @RequestParam(value = "iac", required = false) final boolean iac) {
-    log.with("party_id", partyId).debug("Retrieving cases by party");
-    List<Case> casesList = caseService.findCasesByPartyId(partyId, iac);
+      @RequestParam(value = "iac", required = false) final boolean iac,
+      @RequestParam(value = "max_cases_per_survey", required = false)
+          final Integer maxCasesPerSurvey) {
+
+    List<Case> casesList;
+    if (maxCasesPerSurvey != null) {
+      log.with("party_id", partyId)
```
mfcrocker: This is an example of good logging and why we don't need the rest of the noise

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
AndrewTorrance: Agree  amman , changed to similar to that, but followed the existing convention of 'entering <func name>' . In this case its better than 'Finding cases by party id' since we have two methods of doing that, one not limited by survey , and the new one that does,

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
AndrewTorrance: Changed

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
AndrewTorrance: Implemented a local cache with a hash map

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
AndrewTorrance: Was removed

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
AndrewTorrance: Fair catch , changed

PR #131, opened 2020-05-01
Removing ONS dependencies

akssinhaamit:
# Motivation and Context

Removing ONS dependencies to make the service independent.
# What has changed

pom.xml

yes

unit and IT
# How to test?

# Links


akssinhaamit:
Not sure why code cov is not working...

PR #138, opened 2020-06-16
create json marshalling bean

jterryons:
# Motivation and Context

# What has changed

# How to test?

# Links


src/main/java/uk/gov/ons/ctp/response/casesvc/message/ReceiptFilter.java
```diff
@@ -0,0 +1,13 @@
+package uk.gov.ons.ctp.response.casesvc.message;
+
+import org.springframework.integration.core.MessageSelector;
+import org.springframework.messaging.Message;
+
+public class ReceiptFilter implements MessageSelector {
+
+    @Override
+    public boolean accept(Message<?> message) {
+        //how do we validate this?
```
mfcrocker, line 10: Does this need implementing? Changing this to not validate rather than validating against the schema feels like a regression to me

Jon-Kent: Agreed, I've added basic validation.

PR #143, opened 2020-06-29
Remove the /cases endpoint

warrenbailey:
# Motivation and Context

Removes the /cases endpoint as it is not used by any other services and exposes a risk as calling it with no parameters returns the entire case table.

# What has changed

# How to test?

# Links


README.md
```diff
@@ -104,7 +104,7 @@ There are two ways of running this service
 To use the code styler please goto this url (https://github.com/google/google-java-format) and follow the Intellij instructions or Eclipse depending on what you use
 
 ## API
-Open API spec can be found [here](API.yaml)
+Open API spec can be found~~~~ [here](API.yaml)
```
insacuri, line 107: Accidental change?

src/main/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseEndpoint.java
```diff
@@ -190,38 +190,6 @@ public CaseEndpoint(
...
-  private List<Case> getCases(Example<Case> exampleCase) {
-    boolean emptyRequest = exampleCase.getProbe().equals(new Case());
-    if (emptyRequest) {
-      return caseRepository.findAll();
-    } else {
-      return caseRepository.findAll(exampleCase);
-    }
-  }
-
   private Example<Case> buildExampleCase(String sampleUnitId, String partyId) {
```
insacuri, line 193: This can be deleted too as the only thing that used it is in the deleted block above

README.md
warrenbailey: Yes!

src/main/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseEndpoint.java
warrenbailey: I thought I'd done that, good spot!

PR #147, opened 2020-07-13
removed XMLGregorianCalendar from receipts

Jon-Kent:
# Motivation and Context
Removed XMLGregorianCalendar from receipts.
Instead of being sent a time in the json payload and converting it to a calendar, case service is now
responsible for creating the Timestamp.


src/main/java/uk/gov/ons/ctp/response/casesvc/message/CaseReceiptReceiver.java
```diff
@@ -34,6 +36,8 @@
 
   @Autowired private CaseService caseService;
 
+  @Autowired private TimeHelper timeHelper;
```
mfcrocker, line 39: See comment below but I don't feel like we should be instantiating one of these

src/main/java/uk/gov/ons/ctp/response/lib/common/time/DateTimeUtil.java
```diff
@@ -31,6 +31,15 @@ public static Timestamp nowUTC() {
     return new Timestamp(System.currentTimeMillis());
   }
 
+  /**
+   * Non static method to retrieve the current time as a Timestamp.
+   *
+   * @return The current time in UTC
+   */
+  public Timestamp getNowUTC() {
```
mfcrocker, line 39: I don't quite get why this method has been implemented as non-static; this is a library class, it makes sense to call it in a static manner.

(Honestly, having a method just for `Timestamp(System.currentTimeMillis());` is weird but whatever)

src/main/java/uk/gov/ons/ctp/response/lib/common/time/TimeHelper.java
```diff
@@ -0,0 +1,19 @@
...
+import java.sql.Timestamp;
+
+/**
+ * This interface is to provide a layer of separation
+ * between the service classes and DateTimeUtil.
+ *
+ * This will aid us in incrementally removing the
+ * XMLGregorianCalendar from the services.
+ */
+public interface TimeHelper {
```
mfcrocker, line 12: I don't think this is a useful abstraction; we can just remove instances of `DateTimeUtil.giveMeCalendarForNow();` and replace them with `DateTimeUtil.nowUTC();`, I don't think we're ever going to make a different implementation of `TimeHelper` that we switch over to

src/main/java/uk/gov/ons/ctp/response/lib/common/time/DateTimeUtil.java
Jon-Kent: Removing this code from the CaseReceiptReceiver class and placing it in a non static method allows its returned value to be easily controlled (via mocking the DatetimeUtil class) within tests. It is created as a spring singleton so the overhead is negligible. The timestamp was previously passed in to the service, meaning that the tests could simply create the time, pass it to the service, and then check for that same time in the output. Now that the timestamp is created by the service, the tests can no longer control it unless they are able to mock out the code responsible for creating it - thus its placement in a non static method.

src/main/java/uk/gov/ons/ctp/response/lib/common/time/TimeHelper.java
Jon-Kent: I have removed the TimeHelper interface for now to keep things simple. However, I disagree that we will not make different implementations as that is exactly what we should be doing for integration testing.

PR #162, opened 2020-12-10
Active enrolment enable work

akssinhaamit:
# Motivation and Context

As per the new design described [here](https://github.com/ONSdigital/ras-rm-documentation/tree/master/service-design/improvements/action-service-restructure-design), this PR makes case to use active enrolment instead of actionPlanId which is behind the feature flag.

# How to test?
Smoke Test

# Links

[Trello](https://trello.com/c/jpQWWct4/246-s24-use-active-enrolments-instead-of-action-plan)


src/main/java/uk/gov/ons/ctp/response/casesvc/definition/CaseCreation.java
```diff
@@ -57,9 +59,9 @@
 
   @XmlElement(required = true)
   protected String sampleUnitType;
-
-  @XmlElement(required = true)
+  @XmlElement(nillable = true)
```
insacuri, line 62: Can you tidy up the spacing?  It's a bit hard to read that actionplanid is nillable.  There should also be a space below that for the activeenrolment variable too

src/main/java/uk/gov/ons/ctp/response/casesvc/domain/model/Case.java
```diff
@@ -88,9 +88,12 @@
   @Column(name = "collectioninstrumentid")
   private UUID collectionInstrumentId;
 
-  @Column(name = "actionplanid")
+  @Column(name = "actionplanid", nullable= true)
   private UUID actionPlanId;
 
+  @Column(name = "active_enrolment")
```
insacuri, line 94: Does this element also need to be nullable?  What happens if this isn't included (like it is in the current version with the actionplanid)

src/main/java/uk/gov/ons/ctp/response/casesvc/message/notification/CaseNotification.java
```diff
@@ -41,10 +41,11 @@
 
   @XmlElement(required = true)
   protected String caseId;
-
-  @XmlElement(required = true)
+  @XmlElement(nillable = true)
```
insacuri, line 44: Add a space above this to make it easier to read

src/main/resources/application.yml
```diff
@@ -92,6 +92,7 @@ internet-access-code-svc:
     password: secret
 
 action-svc:
+  deprecated: true
```
insacuri, line 95: the default for now should be false

src/main/resources/logback.xml
```diff
@@ -65,7 +65,7 @@
       </providers>
     </encoder>
     <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
-      <level>DEBUG</level>
+      <level>INFO</level>
```
insacuri, line 68: Why was this changed?  This'll make it so that case only logs info lines right?  If so, we want it to log at debug level so we can have more and filter out what we don't need

src/test/java/uk/gov/ons/ctp/response/casesvc/CaseCreator.java
```diff
@@ -44,7 +44,7 @@ public CaseNotification sendSampleUnit(
     SampleUnitParent sampleUnit = new SampleUnitParent();
     sampleUnit.setCollectionExerciseId(collectionExerciseId.toString());
     sampleUnit.setId(sampleUnitId.toString());
-    sampleUnit.setActionPlanId(UUID.randomUUID().toString());
+    //sampleUnit.setActionPlanId(UUID.randomUUID().toString());
```
insacuri, line 47: Remove if not needed

src/test/resources/application-test.yml
```diff
@@ -12,6 +12,7 @@ survey-svc:
     retry-attempts: 5
 
 action-svc:
+  deprecated: true
```
insacuri, line 15: I think this should be false?  Not sure

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
```diff
@@ -432,23 +439,34 @@ private void processActionPlanChange(final Case targetCase, final boolean enrolm
...
+          }
+          for (Case caze : cases) {
+            if (caze.getSampleUnitType() == SampleUnitType.B) {
+              caze.setActionPlanId(actionPlans.get(0).getId());
+              caseRepo.saveAndFlush(caze);
+              notificationPublisher.sendNotification(
+                      prepareCaseNotification(caze, CaseDTO.CaseEvent.ACTIONPLAN_CHANGED));
+            }
+          }
+        }else {
```
insacuri, line 462: spacing here is wrong

src/main/java/uk/gov/ons/ctp/response/casesvc/domain/model/Case.java
akssinhaamit: Should default to false as its boolen

src/main/resources/application.yml
akssinhaamit: 👍 .

src/main/resources/logback.xml
akssinhaamit: It is on DEBUG level which makes the log size really big on GCP for devs. We should only make it debug when it is required. It was wrong so changed it to info

src/test/java/uk/gov/ons/ctp/response/casesvc/CaseCreator.java
akssinhaamit: Sure.. But If I remove it I can't mark the test config deprecated as true. I will uncomment it for now.

src/test/resources/application-test.yml
akssinhaamit: 👍

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseService.java
akssinhaamit: 👍

src/main/java/uk/gov/ons/ctp/response/casesvc/definition/CaseCreation.java
akssinhaamit: 👍

src/main/java/uk/gov/ons/ctp/response/casesvc/message/notification/CaseNotification.java
akssinhaamit: 👍

src/test/java/uk/gov/ons/ctp/response/casesvc/CaseCreator.java
akssinhaamit: 👍

jterryons (approved):

PR #167, opened 2020-12-22
Fix the issue with frontstage failure

akssinhaamit:
# Motivation and Context
While creating the active-enrolment field the default constraint was not set due to which when the Front stage looks for old cases it fails in preprod with null value.
# What has changed

# How to test?

# Links

### DIFF
--- a/src/main/resources/database/changelog-master.yml
+++ b/src/main/resources/database/changelog-master.yml
@@ -109,3 +109,6 @@ databaseChangeLog:
 
   - include:
       file: database/changes/release-24/changelog.yml
+
+  - include:
+      file: database/changes/release-24/changelog.yml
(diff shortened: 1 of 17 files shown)


src/main/resources/database/changelog-master.yml
```diff
@@ -107,5 +107,8 @@ databaseChangeLog:
   - include:
       file: database/changes/release-23/changelog.yml
 
+  - include:
+      file: database/changes/release-24/changelog.yml
```
jterryons, line 111: release-25

jterryons (changes requested):

src/main/resources/database/changelog-master.yml
akssinhaamit: done

akssinhaamit: the next one would be release-25

insacuri (approved):

PR #206, opened 2021-07-02
Remove the responses table

mfcrocker:
# What and why?
The responses table isn't actually used; we store it and return it in the API, but the only service that cares about it (rm-sdx-gateway) hard-codes the response to OFFLINE so the data is completely obsolete.

This gets rid of it in order to make later refactoring easier

# How to test?

# Trello
https://trello.com/c/LpvQgqX3/32-remove-responses-table-in-rm-case-service-8


warrenbailey:
LGTM however as you're removing all the responses are we able to get rid of this now: https://github.com/ONSdigital/rm-case-service/blob/42115a569e616af62547b34a39f26bc1e62e1316/src/main/java/uk/gov/ons/ctp/response/casesvc/message/CaseReceiptPubSubSubscription.java#L57

warrenbailey (approved):

PR #211, opened 2021-07-23
Remove unused categories

insacuri:
# What and why?

There are a number of categories that apply to  social and census that we don't support.  This PR removes a number of them (though not all of them)

# How to test?

# Trello


creativeappdevamit (approved):
I have no knowledge of which category are not in use but the PR looks ok to me with respect to code changes :-).

PR #245, opened 2022-02-14
making changes to support multiple templates based on form types

creativeappdevamit:
# What and why?

# How to test?

# Trello
https://trello.com/c/A8u90DZF/1344-make-changes-to-case-service-to-pass-form-type-as-classifiers-while-giving-a-call-to-the-notify-gateway

### DIFF
--- a/_infra/helm/case/values.yaml
+++ b/_infra/helm/case/values.yaml
@@ -87,5 +87,5 @@ crons:
     target: "retry-event"
 
 surveySvc:
-  multipleFormTypeSupportedSurveysIds: "202"
+  multipleFormTypeSupportedSurveyIds: "202"
   multipleFormTypeSupported: "1862,1864,1874"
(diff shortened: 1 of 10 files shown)


_infra/helm/case/values.yaml
```diff
@@ -85,3 +85,7 @@ crons:
     name: case-action-event-retry-scheduler
     cron: "0 * * * *"
     target: "retry-event"
+
+surveySvc:
+  multipleFormTypeSupportedSurveysIds: "202"
```
insacuri, line 90: 2 things:
- multipleFormTypeSupportedSurveysIds should be multipleFormTypeSupportedSurveyIds, as it's more grammatically correct.  If you take on this change, you'll have to make the change everywhere else that uses that config item name!
- Is there a reason why this one is only 202 and the one in application.yml has 202 and 203?

creativeappdevamit: No reason, just a guidance that it's comma separated . With respect to naming I will change it.

insacuri (approved):
good to merge as soon as the variable name change is done!

PR #247, opened 2022-03-10
fuel survey support for template

creativeappdevamit:
# What and why?
Case service changes to support for quarterly fuels survey support
# How to test?

# Trello
https://trello.com/c/mzt8q6oH/1397-quarterly-fuels-survey-multiple-template-support

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -112,6 +112,7 @@ survey-svc:
     read-timeout-milli-seconds: 5000
   multiple-form-type-supported-survey-ids: "202,203"
   multiple-form-type-supported: "1862,1864,1874"
+  survey-id-supported-template: "024,283"
 
 case-distribution:
   retrieval-max: 50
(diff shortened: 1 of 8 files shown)


src/main/resources/application.yml
```diff
@@ -143,13 +144,13 @@ retries:
   backoff: 5000
 
 gcp:
-  project: ras-rm-sandbox
-  receiptSubscription: ras-rm-receipt
-  caseNotificationTopic: "test_topic"
-  caseNotificationSubscription: "test_subscription"
+  project: ras-rm-dev
+  receiptSubscription: sdx-receipt-sinhaa
```
insacuri, line 148: Probably wanna revert these before you merge it in

insacuri (changes requested):
Code looks good.  Just need to revert the config changes and it's good to go

src/main/resources/application.yml
creativeappdevamit: ahh... sorry for this.. will revert it

PR #248, opened 2022-03-30
move dockerfile into _infra folder

anwilkie:
# What and why?
The dockerfile for Java repos needed to be moved from the root to the _infra/docker folder so that it's like the Python repos

# Trello
[Card](https://trello.com/c/lt82nJM8)


.github/workflows/main.yml
```diff
@@ -70,7 +70,7 @@ jobs:
       - name: Build Docker Image
         if: github.ref != 'refs/heads/main'
         run: |
-          docker build -t "$REGISTRY_HOSTNAME"/"$HOST"/"$IMAGE":${{ steps.tag.outputs.pr_number }} .
+          docker build -t "$REGISTRY_HOSTNAME"/"$HOST"/"$IMAGE":${{ steps.tag.outputs.pr_number }} -f _infra/docker/Dockerfile .
```
insacuri, line 73: super minor, but can you have both changes follow the same pattern, one is -f -t, the other is -t -f

PR #251, opened 2022-06-06
bump jackson version to latest version

JacobHarrisonONS:
# What and why?
bump jackson version to latest version fixing vulnerability 

# How to test?

# Trello


LJBabbage (approved):
Tested locally and then on GCP, case generation completed as expected

PR #257, opened 2022-07-25
bump springboot and dependents

JacobHarrisonONS:
# What and why?
Looks like springfox/swagger had to go, kept breaking. 

# How to test?

# Trello


insacuri (approved):
Tested it in a few situations in dev and it seems happy.  When there isn't a schema it errors on startup but that's the same as all the other java repos we've updated so nothing to worry about there.

PR #266, opened 2023-02-24
Add valid transition for valid view action

insacuri:
# What and why?

# How to test?

# Trello


src/main/java/uk/gov/ons/ctp/response/casesvc/state/CaseSvcStateTransitionManagerFactory.java
```diff
@@ -263,6 +263,16 @@ private StateTransitionManager<CaseState, CaseEvent> caseStateTransitionManager(
         CategoryDTO.CategoryName.COMPLETED_TO_NOTSTARTED,
         CaseGroupStatus.NOTSTARTED);
 
+    // In response-operations-ui, looking at the cases for a survey also shows you an unused IAC
```
LJBabbage, line 266: The formatting is a bit odd

LJBabbage (approved):
All good, the comment formatting is a bit strange, but I tried before and after and it is removed and makes sense

src/main/java/uk/gov/ons/ctp/response/casesvc/state/CaseSvcStateTransitionManagerFactory.java
insacuri: Blame the maven formatter for that >_<

PR #269, opened 2023-04-04
Add delete by ce endpoint

JacobHarrisonONS:
# What and why?
With the broken BRES collection exercise, we needed a way to delete the unusable cases in the database.
EDIT:
Modified PR to now work with a cascade on FKs from  casegroup allowing the ability to directly delete from collection exercise id hopefully allowing the sql server to carry the load and not overload case service making it unusual if a large collection was deleted during business hours. 
Corrected url so collection exercise isn't plural

# How to test?
Deploy this to dev, run acceptance tests to get quick case to delete or create one yourself. 
Go to database and find your case, Case Group table easy place to find collection exercise id, you will want to open Case and Case Event since you will be checking these later.
Using postman DELETE with http://localhost:8171/casegroups/collectionExercise/{collection_excercise_id}
Check database tables (Case Group, Case and Case Event) to see that case was deleted and nothing was left associated with that case and log to ensure no errors (there shouldn't be errors).
Check outputs and logs for a collection that exists and doesn't exist. 

# Jira
https://jira.ons.gov.uk/browse/RAS-283


src/main/java/uk/gov/ons/ctp/response/casesvc/domain/repository/CaseGroupRepository.java
```diff
@@ -70,23 +70,12 @@ Long findCasesAgainstCollectionExerciseID(
       @Param("collectionExerciseId") UUID collectionExerciseId);
 
   /**
-   * find cases for action
+   * find case groups for a collection exercise
    *
    * @param collectionExerciseId the collection exercise id
    * @return the list of case actions
    */
-  @Query(
```
JacobHarrisonONS, line 78: this didnt get removed, just moved down, git is being funny

pricem14pc (changes requested):
We should add the new `DELETE` method to the `/cases` endpoint in the `openapi.yaml`

src/main/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseEndpoint.java
```diff
@@ -381,4 +376,31 @@ private CaseDetailsDTO buildDetailedCaseDTO(Case caze, boolean caseevents, boole
...
     return caseDetailsDTO;
   }
+
+  /**
+   * Deletes all case, caseevent and casegroup data for a particular collection exercise
+   *
+   * @param collectionExerciseId The Collection Exercise UUID to delete for
+   * @return An appropriate HTTP repsonse code
+   */
+  @DeleteMapping("{collectionExerciseId}")
```
pricem14pc, line 386: Its unfortunate that the existing API is a bit confusing, but I don't think this new endpoint is quite right.

There is a `GET` method is defined as  `/cases/{caseId}`. I don't think the equivalent `DELETE` method for that resource should accept a different parameter (i.e. the `collectionExerciseId`).

Furthermore, the collection exercise is an attribute of the casegroup and not the case. Therefore the request should probably be against the parent `/casegroup` resource.

There is already a `GET` method defined against the `casegroup` resource to retrieve a count of casegroups by `collectionExerciseId` so it would be more appropriate to have the `DELETE` method act on that.

However, it may be even more appropriate again to have the `DELETE` act on a new route called `/casegroups/collectionExercises/{collectionExerciseId}`

pricem14pc (changes requested):

pricem14pc:
Using the Hibernate JPA object mapping in this way has the consequence of Hibernate making separate sequential SQL delete requests for each record (in each table) by foreign key.

While this approach may be ok for small samples, it is not scalable and should not be integrated with synchronous UI actions (particularly for those collection exercises that have been live and generated case events).

At the moment the largest collection exercise has ~40k cases and ~80k case events.. This would result in Hibernate making 250k separate delete requests by primary/foreign keys. In addition there are expensive select statements to identify the records to be deleted.

pricem14pc (changes requested):

src/main/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseEndpoint.java
```diff
@@ -381,4 +376,31 @@ private CaseDetailsDTO buildDetailedCaseDTO(Case caze, boolean caseevents, boole
...
+    List<Integer> listOfCasePKs = caseList.stream().map(Case::getCasePK).toList();
+    List<CaseEvent> caseEventsList = caseEventRepository.findByCaseFKIn(listOfCasePKs);
+
+    caseEventRepository.deleteAll(caseEventsList);
+    caseService.deleteCasesInList(caseList);
+    caseGroupService.deleteCaseGroups(caseGroupList);
+
+    log.with("collection_exercise_id", collectionExerciseId).info("Delete successful");
+
+    return ResponseEntity.ok("Deleted Successfully");
```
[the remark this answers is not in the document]
JacobHarrisonONS: Now log with a 404 when the result of count for casegroups for collection exercise id returns 0.

src/main/java/uk/gov/ons/ctp/response/casesvc/domain/repository/CaseGroupRepository.java
```diff
@@ -147,4 +137,14 @@ List<CaseAction> findByCollectionExerciseIdAndActiveEnrolment(
               + "WHERE cg.caseGroupPK=c.caseGroupFK AND c.state='ACTIONABLE' "
               + "AND c.id IN :caseIds ")
   List<CaseAction> findByCaseIdIn(@Param("caseIds") List<UUID> caseIds);
+
+  /**
+   * find cases for action
+   *
+   * @param collectionExerciseID UUID of the collection exercise to delete casegroup entities
+   */
```
JacobHarrisonONS, line 145: using the modifying and query annotations returns an int of deleted rows, the inbuilt delete returns void.

src/main/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseGroupEndpoint.java
```diff
@@ -154,4 +152,36 @@ private boolean isValidEvent(String event) {
...
+    Long caseGroupCount =
+        caseGroupService.getAllCasesAgainstCollectionExerciseId(collectionExerciseId);
+    if (caseGroupCount == 0) {
+      log.with("collection_exercise_id", collectionExerciseId)
+          .info("No content found for Collection Exercise ID");
+      return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+    }
+
+    log.with("collection_exercise_id", collectionExerciseId)
+        .info("Deleting cases {}", caseGroupCount);
```
pricem14pc, line 177: Is the `caseGroupCount` actually a `caseCount`. If we do this I think we should ignore counting the cases and simply count the casegroups (as that's what we're deleting). We know the database relationship integrity will take care of case and caseevent records.

However, regardless of that, do we need to do the above count at all, or simply use the response from the `deleteCaseGroupByCollectionExerciseId` request as to whether any data was deleted? (i.e. we should get a response of `0` from the deletion request). You could argue what's the point in making the delete request if there's no data to delete but it won't be much more expensive than doing a count up front.

pricem14pc (changes requested):

src/main/java/uk/gov/ons/ctp/response/casesvc/endpoint/CaseGroupEndpoint.java
```diff
@@ -162,26 +168,30 @@ private boolean isValidEvent(String event) {
...
     log.with("collection_exercise_id", collectionExerciseId)
-        .info("Deleting cases {}", caseGroupCount);
+        .info("Deleting cases");
 
     int deletedRows = caseGroupService.deleteCaseGroupByCollectionExerciseId(collectionExerciseId);
 
     log.with("collection_exercise_id", collectionExerciseId)
-        .info("Deleted {} of {} expected cases successfully", caseGroupCount, deletedRows);
-    String output = "Deleted " + deletedRows + " cases successfully";
+        .info("Deleted {} expected cases successfully", deletedRows);
```
pricem14pc, line 176: `.info("Deleted {} expected casegroups successfully", deletedRows);`

pricem14pc (approved):
Reviewed and successfully functionally tested in a low data volume dev environment. As discussed, this endpoint will not be integrated with any other services until it's been tested in a volume environment.

PR #278, opened 2023-09-07
standardise hibernte settings

JacobHarrisonONS:
# What and why?
Align hibernate settings with what works on other services

Testing:
In Dev booted correctly when older version existed with and without data
In Dev booted correctly with only the schema existing 

# How to test?
Test x 2 through Docker and Cloud SQL environment

For both docker and Cloud SQL test x 2 through new env and data retention env

# Jira
https://jira.ons.gov.uk/browse/RAS-801


src/main/resources/application.yml
```diff
@@ -57,17 +64,10 @@ spring:
...
     url: jdbc:postgresql://localhost:5432/ras
-    liquibase-schema: casesvc
     change-log: classpath:/database/changelog-master.yml
+    liquibase-schema: casesvc
+    default-schema: casesvc
     enabled: false
-  main:
-    allow-bean-definition-overriding: true
-  output:
-    ansi:
```
JacobHarrisonONS, line 66: i moved this

src/main/resources/application.yml
```diff
@@ -41,14 +49,13 @@ spring:
...
     defer-datasource-initialization: true
     hibernate:
-      ddl-auto: update
+      ddl-auto: none
       default_schema: casesvc
     properties:
       hibernate:
-        hbm2ddl:
-          schema_filter_provider: uk.gov.ons.ctp.response.casesvc.CustomProvider
-        id:
```
JacobHarrisonONS, line 50: This Custom provider originally thought to do something is not needed as far as I know (it has been removed from sample and collection exercise)

pricem14pc: That's correct, it was removed from sample in favour of Liquibase handling migrations. The only reason this had been introduced was an attempt to stop Hibernate from creating certain objects. That did not work as it does not honour the `dd-auto: update` strategy.

src/main/resources/application.yml
pricem14pc: Agreed, makes sense

pricem14pc (approved):
Tested all 4 scenarios
Cloud SQL new env and existing env
Docker Postgres new env and existing env
All behaving as expected with Liquibase handling migrations rather than hibernate

PR #280, opened 2023-09-20
rename NOT_ENGLAND variable to NI_REGION_CODE 

JacobHarrisonONS:
# What and why?
Fix a bit of confusion around the following bit of code 

{{ // if not not england is england. welcome to ras-rm.
if (!NOT_ENGLAND.equals(region)) {
region = "";
}}}

rename NOT_ENGLAND so it can make more sense

# How to test?

# Jira

https://jira.ons.gov.uk/browse/RAS-769

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/casesvc/representation/action/NotifyModel.java
+++ b/src/main/java/uk/gov/ons/ctp/response/casesvc/representation/action/NotifyModel.java
@@ -98,7 +98,7 @@ public class NotifyModel {
         private static final String REMINDER = "REMINDER";
         private static final String NOTIFICATION = "NOTIFICATION";
         // a region of YY appears to be any region that isnt england
-        private static final String NOT_ENGLAND = "YY";
+        private static final String NI_REGION_CODE = "YY";
 
         private String actionType;
         private String legalBasis;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/ons/ctp/response/casesvc/representation/action/NotifyModel.java
```diff
@@ -115,8 +115,8 @@ public Classifiers build() {
             return new Classifiers(actionType, "", "", "", "");
           }
 
-          // if not not england is england. welcome to ras-rm.
-          if (!NOT_ENGLAND.equals(region)) {
+          // if NOT NI then England
```
SteveScorfield, line 118: I'd put England/Wales just to be 100% clear

SteveScorfield (approved):
Looks good. One minor alteration the comments but that's it.

src/main/java/uk/gov/ons/ctp/response/casesvc/representation/action/NotifyModel.java
JacobHarrisonONS: okay - ideally we wouldnt need the comment but its already been confusing

SteveScorfield: Agreed

PR #282, opened 2023-11-14
Migrate from openjdk to eclipse-temurin

pricem14pc:
# What and why?

The openjdk base images are deprecated and no longer maintained. This migrates to the recommended [eclipse-temurin](https://hub.docker.com/_/eclipse-temurin) image. This also migrates from the slim Linux OS to the alpine OS images as eclipse-temurin doesn't appear to offer slim. This reduces the image from ~305BM to ~143MB

# How to test

Does the image build, does the application start, and do the acceptance tests pass?

### DIFF
--- a/_infra/docker/Dockerfile
+++ b/_infra/docker/Dockerfile
@@ -1,11 +1,9 @@
-FROM openjdk:17-slim
+FROM eclipse-temurin:17-jre-alpine
 
 ARG JAR_FILE=case.jar
-RUN apt-get update
 
 ENV JAVA_OPTS=""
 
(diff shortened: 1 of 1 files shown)


_infra/docker/Dockerfile
```diff
@@ -1,11 +1,9 @@
-FROM openjdk:17-slim
+FROM eclipse-temurin:17-jre-alpine
 
 ARG JAR_FILE=case.jar
-RUN apt-get update
```
pricem14pc, line 4: `apt-get` isn't installed in alpine, but we don't install anything in this image anyway.

PR #284, opened 2023-11-30
Update Mockito-inline to Mockito-core

matthew-robinson-ons:
# What and why?
This PR updates mockito-inline to mockito-core and adds the required byte-buddy dependency.

GHA auth deprecation has also been addressed in this PR. 

# How to test?
- Run local tests 
- Deploy and run acceptance tests
- 
# Jira
https://jira.ons.gov.uk/secure/RapidBoard.jspa?rapidView=1493&view=detail&selectedIssue=RAS-949

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -274,9 +274,14 @@
         </dependency>
         <dependency>
             <groupId>org.mockito</groupId>
-            <artifactId>mockito-inline</artifactId>
-            <version>4.6.1</version>
+            <artifactId>mockito-core</artifactId>
+            <version>5.5.0</version>
             <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>net.bytebuddy</groupId>
+            <artifactId>byte-buddy</artifactId>
+            <version>1.14.8</version>
         </dependency>
         <dependency>
             <groupId>javax.xml.ws</groupId>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -274,10 +274,15 @@
...
-            <artifactId>mockito-inline</artifactId>
-            <version>4.6.1</version>
+            <artifactId>mockito-core</artifactId>
+            <version>5.5.0</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>net.bytebuddy</groupId>
+            <artifactId>byte-buddy</artifactId>
+            <version>1.14.8</version>
```
arroyoAle, line 284: While we're here can we update them to the latest version on maven?

arroyoAle (changes requested):
tests run locally and in GHA, app works in GCP and acceptance tests pas, only comment is there are newer versions of the packages we are implementing

pom.xml
matthew-robinson-ons: Initially left them the same as previous 2 service changes for consistency. Will retest on latest version as well.

arroyoAle: Yeah that makes sense, I'm not sure which way we want to go...

arroyoAle (approved):
Happy with this. Also see comment above

LJBabbage (approved):

PR #295, opened 2024-10-14
Initial investigation of Java 21 spring boot 3

pricem14pc:
Superseded by https://github.com/ONSdigital/rm-case-service/pull/303

# What and why?

Initial sanity test of upgrading Java and Spring Boot

- Upgrade to Java 21
- Upgrade to Spring Boot 3

## Outstanding Issues

The app now starts and successfully amd connects to Postgres and Pub/Sub. The app successfully deploys to GCP and acceptance tests all pass successfully.

For current progress/outstanding issues check the `TODO:` comments throughout for more details. In summary so far.

- failsafe integration tests currently disabled due to the Spring test server rendering an internal server error HTTP response. **UPDATE**: This was due to incompatible godaddy libraries.
```
{"error":
   {"code":"SYSTEM_ERROR",
    "timestamp":"20241101092339701",
    "message":"Could not write JSON: Class versions V1_5 or less must use F_NEW frames."}
}
```

We have to update WireMock from the legacy `com.github.tomakehurst` to the new `org.wiremock` library. This new hosted package has the original tomakehurst package classes bundled inside it. This means we don't need to update the various imports throughout as they've made usage backward compatible.

However, we see ReflectASM errors with the GoDaddy logging framework (only when using the new `org.wiremock` package, not the legacy `com.github.tomakehurst package`).

> Class versions V1_5 or less must use F_NEW frames

The RefectASM and GoDaddy logging packages are both very old and no longer supported https://github.com/godaddy/godaddy-logger/blob/main/README.md#whats-included-in-this-library.

The errors themselves maybe because of the way they've bundled the original tomakehurst within the new wiremock library, and the reflection framework is unable to instrument the byte code correctly under Java 21. **UPDATE**: The godaddy and wiremock errors turned out to be completely unrelated.


pom.xml
```diff
@@ -372,21 +385,37 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
-                <version>3.12.1</version>
+                <version>3.8.1</version>
                 <configuration>
                     <source>${java.version}</source>
                     <target>${java.version}</target>
+                    <parameters>true</parameters>
```
pricem14pc, line 392: Required to fix failing `CaseGroupEndpointUnitTest` tests.
>Error "Name for argument of type [java.util.UUID] not specified, and parameter name information not found in class file either" with spring 3.2 upgrade when using `org.springframework.test.web.servlet.MockMvc`

pom.xml
```diff
@@ -11,14 +11,15 @@
...
         <common.version>10.49.4</common.version>
         <jackson.version>2.16.1</jackson.version>
-        <java.version>17</java.version>
+        <java.version>21</java.version>
         <surefire.version>3.2.5</surefire.version>
-        <godaddylogging.version>1.2.5</godaddylogging.version>
+        <!-- godaddylogging.version>1.2.5</godaddylogging.version -->
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+        <failsafe.skipITs>true</failsafe.skipITs>
```
pricem14pc, line 22: To allow me to control the tests during development. This will of course be `false` when delivered.

pricem14pc:
https://github.com/ONSdigital/rm-case-service/pull/303

PR #296, opened 2024-12-27
gar migration

anwilkie:
# What and why?
Migrating case service from GCR to GAR.

# How to test?
Deploy this to your environment, setting configBranch to `ras-1405-case-gar-migration`. The deployment will hang until you replace the image in the YAML with `europe-west2-docker.pkg.dev/ons-ci-rmrasbs/images/case:pr-296`.

# Jira
[Card](https://jira.ons.gov.uk/browse/RAS-1405)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @ONSdigital/sdc-rmras
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,9 +14,10 @@ on:
 
 env:
   IMAGE: case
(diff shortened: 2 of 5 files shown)


LJBabbage (approved):
This does work, but it's not how it should be tested. By deploying with spinnaker or from  the terminal the yaml changes will not be reflected and therefore can not be verified. You should update the env, namespace and tag in the yaml, kill the current templates in flight (service, deployment, cron etc) then deploy everything through HELM. The config branch is nothing but a replacement for values and when you have new vars/names in the yaml it won't be tested properly

matthew-robinson-ons (approved):
Looks good. We should add a CODEOWNERS file in this repo

PR #300, opened 2025-03-05
Adding graceful handling of duplicate case creation

pricem14pc:
# What and why?

For samples over a certain size, we periodically observe case creation messages being redelivered during setting ready for live. This appears to be due to the case service not being able to process a single message quickly enough, and the redelivery for unacknowledged messages happens while cases are being created.

This results in cases in the database, and then subsequent requests for the same case, causing constraint violations, which themselves result in no ack and redelivery of the same message multiple times until deadline.

Compounding this is the Spring error exception around unexpected rollback (within a nested transaction), that buries the underlying Postgres constraint violation being raised.

The safety net being introduced by this PR means we don't simply throw the Spring rollback exception, but now handle the constraint violation, raise our own new `DUPLICATE_RECORD` CTPException and ack the message. This is safe as the only reason the message exists is to create the record. While not ideal, the issue is not exceptional and we can recover from it by acking the message.

We should also try this change in combination with the ack deadline extension now we're not throwing the Spring rollback exception and nacking the message. However, that deadline is arbitrary and dependent on the performance of the case service to handle lots of messages at once so again it isn't the right solution. The right solution is a complete redesign of the case request handler.

# How to test?

Load a large sample and check messages get replayed only once if they're not acknowledged, and then get acked.

# Jira

https://jira.ons.gov.uk/browse/RAS-648

### DIFF
--- a/_infra/helm/case/Chart.yaml
+++ b/_infra/helm/case/Chart.yaml
@@ -14,8 +14,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 12.0.27
+version: 12.0.28
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
-appVersion: 12.0.27
+appVersion: 12.0.28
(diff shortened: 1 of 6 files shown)


mfcrocker (approved):
Looks good code-wise, as discussed I don't think this is reasonably testable other than from a regression standpoint

LJBabbage (approved):

PR #305, opened 2025-10-08
RAS-1107 Create endpoint for casegroup survey, party id

LJBabbage:
# What and why?

# How to test?

# Jira

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/casesvc/domain/repository/CaseGroupRepository.java
+++ b/src/main/java/uk/gov/ons/ctp/response/casesvc/domain/repository/CaseGroupRepository.java
@@ -147,4 +147,15 @@ public interface CaseGroupRepository extends JpaRepository<CaseGroup, Integer> {
   @Query("DELETE FROM CaseGroup cg WHERE cg.collectionExerciseId = :collectionExerciseId")
   int deleteCaseGroupsByCollectionExerciseId(
       @Param("collectionExerciseId") UUID collectionExerciseID);
+
+  /**
+   * To find CaseGroup by party UUID and survey UUID
+   *
+   * @param partyId the UUID of the Party
+   * @param surveyId the UUID of the survey
+   * @return the matching CaseGroups
+   */
+  @Query("SELECT c FROM CaseGroup c WHERE c.partyId = :partyId AND c.surveyId = :surveyId")
+  List<CaseGroup> findByPartyAndSurveyId(
+      @Param("partyId") UUID partyId, @Param("surveyId") UUID surveyId);
 }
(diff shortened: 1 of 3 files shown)


openapi.yaml
```diff
@@ -103,6 +103,42 @@ paths:
...
+        - name: survey_id
+          in: path
+          description: the uuid of the survey
+          required: true
+          schema:
+            type: string
+        - name: limit
+          in: query
+          schema:
+            type: string
```
pricem14pc:
```suggestion
        - name: limit
          in: query
          description: the maximum number of cases to return
          required: true
          schema:
            type: integer
```

src/main/java/uk/gov/ons/ctp/response/casesvc/domain/repository/CaseGroupRepository.java
```diff
@@ -147,4 +149,22 @@ List<CaseAction> findByCollectionExerciseIdAndActiveEnrolment(
...
+  @Query(
+      value =
+          "SELECT new uk.gov.ons.ctp.response.casesvc.representation.ReportingUnitDTO"
+              + "(cg.collectionExerciseId, cg.status, c.id as caseId, c.createdDateTime, iac.iac) "
+              + "FROM CaseGroup cg, Case c "
+              + "LEFT JOIN CaseIacAudit iac ON iac.caseFK=c.casePK "
+              + "WHERE cg.caseGroupPK=c.caseGroupFK AND cg.partyId= :partyId "
+              + "AND cg.surveyId IN :surveyId ORDER BY c.createdDateTime")
+  List<ReportingUnitDTO> findReportingUnitsByPartyAndSurveyId(
+      Pageable pageable, @Param("partyId") UUID partyId, @Param("surveyId") UUID surveyId);
```
pricem14pc: See comment below, potentially we _will_ need to handle pagination in this query but the current use cases don't require it. Would a basic `LIMIT` clause be better? There doesn't seem to be any performance gains between that and the `fetch first 12 rows only` that JPA uses.

pricem14pc (commented):
Couple of minor comments

src/main/java/uk/gov/ons/ctp/response/casesvc/service/CaseGroupService.java
```diff
@@ -85,6 +87,24 @@ public List<CaseGroup> findCaseGroupBySurveyId(final UUID surveyId) {
...
+   * @return list of ReportingUnitDTO
+   */
+  public List<ReportingUnitDTO> getReportingUnitsByPartyAndSurveyId(
+      final UUID partyId, final UUID surveyId, int limit) {
+    log.debug(
+        "Entering findRUDetailsByPartyAndSurveyId",
+        kv("partyId", partyId),
+        kv("survey_id", surveyId));
+    return caseGroupRepo.findReportingUnitsByPartyAndSurveyId(
+        PageRequest.of(0, limit), partyId, surveyId);
```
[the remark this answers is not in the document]
LJBabbage: I was refactoring what we had rather than opening the can of works of the reason why and uses (current or future).

PR #312, opened 2026-02-03
Allow livenessProbe and heap memory configuration

pricem14pc:
# What and why?

An initial attempt to stop case receiving a sigterm during long running processes (mps)

- allow us to override livenessProbe to stop Kubernetes from killing long running processes
- allow explicit `JAVA_OPTIONS` to allocate more heap


matthew-robinson-ons (approved):
Eyeball and previously tested configuration in performance, all looks good .

arroyoAle (approved):
Eyeball review as it is going to be tested in preprod

